PR #1, opened 2024-04-25
update docs & fix CI

dpaiton:
# Resolved Issues
No issues

# Description
- add some minor docstring fixes & moves code to test a PR in the new repo
- fix CI to run lint & tests

### DIFF
--- a/.github/workflows/rust_lint.yml
+++ b/.github/workflows/rust_lint.yml
@@ -20,18 +20,22 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: recursive
+          token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: check out hyperdrive
-        uses: actions/checkout@main
-        with:
-          repository: delvtech/hyperdrive
-          ref: "latest"
-          path: "./hyperdrive"
-
+      # NOTE: This is needed to ensure that hyperdrive-wrappers builds correctly.
       - name: install foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
           version: nightly
+
+      - name: check out hyperdrive
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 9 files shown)


crates/hyperdrive-wrappers/build.rs
```diff
@@ -58,6 +58,12 @@ const TARGETS: &[&str] = &[
 
 fn get_artifacts(artifacts_path: &Path) -> Result<Vec<(String, String)>> {
     let mut artifacts = Vec::new();
+    if !artifacts_path.exists() {
```
jalextowle, line 61: We could just build it here with mkdirp. It’s easy enough that I think it’s worth doing

jalextowle (approved):
LGTM, but we should do a mkdirp

crates/hyperdrive-wrappers/build.rs
dpaiton: oh ok cool, it just needs to be an empty dir? or do we need to populate it with something?

.github/workflows/rust_lint.yml
```diff
@@ -20,18 +20,22 @@ jobs:
...
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
 
       - name: check out hyperdrive
-        uses: actions/checkout@main
+        uses: actions/checkout@v3
         with:
           repository: delvtech/hyperdrive
-          ref: "latest"
```
ryangoree, line 28: What happens if there's no ref? Should we be targeting a specific version?

ryangoree (approved):

.github/workflows/rust_lint.yml
dpaiton: good catch; I pushed up a fix for this to grab use the latest tag.

PR #5, opened 2024-04-26
Various bug fixes in rust sdk

slundqui:
PR originally from https://github.com/delvtech/hyperdrive/pull/1005

# Resolved Issues
This solves various rust crashes resulting from python fuzz testing.
https://github.com/delvtech/hyperdrive/issues/1004

# Description
Fixes the following issues:
- Max long guesses is below the minimum transaction amount, so `calc_open_long` throws a minimum transaction amount error. We clamp the guess to be minimum transaction amount. and do a final check at the end to ensure the max amount is greater than the minimum transaction amount.
- We catch a case in `absolute_max_long` where the `target_share_reserves < effective_share_reserves`. In this case, we throw a better descriptive panic.
- We short circuit `calc_max_long` and return 0 if the spot price after a minimum long exceeds the max spot price.
- Fixing bug with wheel build script.
- Building and uploading to pypi on tag instead of push to main.

# Review Checklists

referenced in the checklist to make sure that they are reviewed. If there are
multiple reviewers, copy the checklists into sections titled `## [Reviewer Name]`.
If the PR doesn't touch Solidity and/or Rust, the corresponding checklist can
be removed.

### Rust

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -318,12 +318,12 @@ impl State {
         let mut maybe_solvency =
             self.solvency_after_short(max_bond_amount, spot_price, checkpoint_exposure);
         if maybe_solvency.is_none() {
-            // If initial guess is insolvant, use the absolute max short
+            // If initial guess is insolvent, use the absolute max short
             // as the initial guess
             max_bond_amount = absolute_max_bond_amount;
             // We expect by definition for the pool to be solvent with the absolute max bond amount.
             maybe_solvency =
-                self.solvency_after_short(max_bond_amount, spot_price, checkpoint_exposure)
+                self.solvency_after_short(max_bond_amount, spot_price, checkpoint_exposure);
         }
         let mut solvency = maybe_solvency.unwrap();
         for _ in 0..maybe_max_iterations.unwrap_or(7) {
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -318,12 +318,7 @@ impl State {
         let mut maybe_solvency =
             self.solvency_after_short(max_bond_amount, spot_price, checkpoint_exposure);
         if maybe_solvency.is_none() {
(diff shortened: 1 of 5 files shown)


crates/hyperdrive-math/src/short/max.rs
```diff
@@ -318,7 +318,7 @@ impl State {
         let mut maybe_solvency =
             self.solvency_after_short(max_bond_amount, spot_price, checkpoint_exposure);
         if maybe_solvency.is_none() {
-            panic!("Initial guess in `max_short` is insolvent.");
+            panic!("Initial guess in `absolute_max_short` is insolvent.");
```
slundqui, line 321: > @[jalextowle](https://github.com/jalextowle) jalextowle [20 hours ago](https://github.com/delvtech/hyperdrive/pull/1005#discussion_r1580134104)
> Didn’t we already check that this isn’t insolvent on line 294? I might be missing something, but I don’t expect this to be solvent
> 
> Member
> Author
> @[slundqui](https://github.com/slundqui) slundqui [19 hours ago](https://github.com/delvtech/hyperdrive/pull/1005#discussion_r1580175165)
> Ah, my mistake, I didn't realize this was in the absolute_max_short function. Is it safe to say here then that if absolute_max_short_guess returns a value that is insolvent, that the pool doesn't allow for a short at all? Or is there a better initial guess we can use here
> 
> Member
> @[jalextowle](https://github.com/jalextowle) jalextowle [14 hours ago](https://github.com/delvtech/hyperdrive/pull/1005#discussion_r1580427013)
> We might be able to use a better initial guess. We could always use 0, or we could allow “cross-overs” by allowing invalid solutions so that we can proceed with iterations. The latter sounds fine to me since Newton’s should get to a solution that is solvent after a few iterations.
> 
> This check was based on my (erroneous) thinking at implementation-time that this algorithm would never cross over the solution

Original conversation from PR, commenting her for visibility

slundqui: Will keep this as is for now, can fix in  follow up PR

dpaiton (approved):

PR #7, opened 2024-04-29
Add `calculate_shares_delta_given_bonds_delta_derivative`

sentilesdal:
# Resolved Issues
  This is a part of adding functions needed to test remove_liquidity

# Description
  - Adds the lower level function `calculate_shares_delta_given_bonds_delta_derivative` to lp.rs
  - Adds a new version of `calculate_present_value_safe` that doesn't panic.  Future work will remove the previous version.
  - Adds a helper function to calculate idle reserves
  - Tests `calculate_shares_delta_given_bonds_delta_derivative`. 

I ran diagnostics to make sure that we were actually testing the function, the test errors on the present value calculation about half the time, errors on the actual function 1/4 time and succeeds about a 1/4 of the time.

# Review Checklists

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/lp.rs
+++ b/crates/hyperdrive-math/src/lp.rs
@@ -1,4 +1,7 @@
-use std::cmp::Ordering;
+use std::{
+    cmp::Ordering,
+    panic::{catch_unwind, AssertUnwindSafe},
+};
 
 use ethers::types::{I256, U256};
 use eyre::{eyre, Result};
@@ -223,6 +226,59 @@ impl State {
         present_value.into()
     }
 
+    // NOTE: This version will not panic.
+    //
+    // Calculates the present value of LPs capital in the pool.
+    pub fn calculate_present_value_safe(
+        &self,
+        current_block_timestamp: U256,
+    ) -> Result<FixedPoint> {
+        let result = catch_unwind(AssertUnwindSafe(|| {
+            // Assume the following methods return Results and can be handled with `?`
+            let long_average_time_remaining = self.calculate_normalized_time_remaining(
+                self.long_average_maturity_time().into(),
+                current_block_timestamp,
+            );
+
+            let short_average_time_remaining = self.calculate_normalized_time_remaining(
+                self.short_average_maturity_time().into(),
+                current_block_timestamp,
+            );
+
+            // Conversion and calculation
+            let share_reserves = I256::try_from(self.share_reserves())
+                .map_err(|_| eyre!("Failed to convert share_reserves"))?;
+
+            let minimum_share_reserves = I256::try_from(self.minimum_share_reserves())
+                .map_err(|_| eyre!("Failed to convert minimum_share_reserves"))?;
+
+            let present_value = share_reserves
+                + self.calculate_net_curve_trade(
+                    long_average_time_remaining,
+                    short_average_time_remaining,
+                )
+                + self.calculate_net_flat_trade(
+                    long_average_time_remaining,
+                    short_average_time_remaining,
+                )
+                - minimum_share_reserves;
+
+            // Check for non-positive present value
+            if present_value < I256::from(0) {
+                Err(eyre!("Negative present value!"))
+            } else {
+                Ok(present_value.into()) // Converting I256 to FixedPoint, assuming it's possible
+            }
+        }));
+
+        // Handling the Result of `catch_unwind`
+        match result {
+            Ok(Ok(present_value)) => Ok(present_value), // Unwrap successful result
+            Ok(Err(e)) => Err(e),                       // Propagate the inner error
+            Err(_) => Err(eyre!("Panic occurred during calculation")),
+        }
+    }
+
     pub fn calculate_net_curve_trade(
         &self,
         long_average_time_remaining: FixedPoint,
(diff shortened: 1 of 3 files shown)


crates/hyperdrive-math/src/lp.rs
```diff
@@ -224,6 +227,61 @@ impl State {
...
+
+            let present_value = share_reserves
+                + self.calculate_net_curve_trade(
+                    long_average_time_remaining,
+                    short_average_time_remaining,
+                )
+                + self.calculate_net_flat_trade(
+                    long_average_time_remaining,
+                    short_average_time_remaining,
+                )
```
jalextowle, line 265: The use of these non-safe functions means that this isn't "safe."

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,122 @@ impl State {
...
+        if derivative < right_hand_side {
+            return Err(eyre!("Derivative is less than right hand side"));
+        }
+        let derivative = derivative - right_hand_side;
+
+        // NOTE: Round up since this is on the rhs of the final subtraction.
+        //
+        // inner = (
+        //             (mu / c) * (k(x) - (y(x) + dy) ** (1 - t_s))
+        //         ) ** (t_s / (1 - t_s))
```
jalextowle, line 486: Looks like the spacing might be messed up here.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,122 @@ impl State {
...
+    }
+
+    /// Given a signed bond amount, this function calculates the negation of the
+    /// derivative of `calculateSharesOutGivenBondsIn` when the bond amount is
+    /// positive or the derivative of `calculateSharesInGivenBondsOut` when the
+    /// bond amount is negative.
+    /// Note that the state is the present state of the pool and original values
+    /// passed in as parameters.  Present sate variables are not expressly
+    /// paased in because so that downstream function like kUp() can still be
+    /// used.
```
jalextowle, line 437: This is missing the Ascii math from the contract code. There is other Ascii math in the function body that is missing. I think it would be best to just duplicate the comments since the documentation is good.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -684,4 +861,107 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn fuzz_test_calculate_shares_delta_given_bonds_delta_derivative() -> Result<()> {
```
jalextowle, line 866: This test is amazing. Really nice job dude!

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,122 @@ impl State {
...
+                return Err(eyre!("Calculating the bond reserves underflows"));
+            }
+        };
+
+        // NOTE: Round up since this is on the rhs of the final subtraction.
+        //
+        // derivative = c * (mu * z_e(x)) ** -t_s +
+        //              (y / z_e) * (y(x)) ** -t_s -
+        //              (y / z_e) * (y(x) + dy) ** -t_s
+        let effective_share_reserves = self.effective_share_reserves();
```
jalextowle, line 463: You're going to want to use a safe variant here.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,122 @@ impl State {
...
+            self.initial_vault_share_price()
+                .mul_down(effective_share_reserves)
+                .pow(self.time_stretch()),
+        ) + original_bond_reserves.div_up(
+            original_effective_share_reserves
+                .mul_down(self.bond_reserves().pow(self.time_stretch())),
+        );
+
+        // NOTE: Rounding this down rounds the subtraction up.
+        let right_hand_side = original_bond_reserves.div_down(
```
jalextowle, line 474: Let's just use the same naming as in the contracts, unless there is something that I'm missing. Switching the style is one thing, but I think it'll be easier to follow if it's exactly the same.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,122 @@ impl State {
...
+        }
+        let derivative = derivative - right_hand_side;
+
+        // NOTE: Round up since this is on the rhs of the final subtraction.
+        //
+        // inner = (
+        //             (mu / c) * (k(x) - (y(x) + dy) ** (1 - t_s))
+        //         ) ** (t_s / (1 - t_s))
+        let k = self.k_up();
+        let y_plus_dy_pow_one_minus_t = bond_reserves_after.pow(fixed!(1e18) - self.time_stretch());
```
jalextowle, line 488: Same here, let's keep the names the same.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,122 @@ impl State {
...
+            )
+        };
+        let derivative = derivative.mul_div_up(inner, self.vault_share_price());
+        let derivative = if fixed!(1e18) > derivative {
+            fixed!(1e18) - derivative
+        } else {
+            return Ok(fixed!(0));
+        };
+        let derivative = if original_share_adjustment >= I256::zero() {
+            let right_hand_side =
```
jalextowle, line 514: Same here, let's use consistent names.

jalextowle (approved):
The impl looks right to me. A few style nits that should be addressed before merging, but otherwise LGTM

crates/hyperdrive-math/src/lp.rs
sentilesdal: thanks yeah you're right, i was just working on a totally safe version actually.  I'll do that in another PR here shortly.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,164 @@ impl State {
             ))
             .unwrap()
     }
+
+    /// Calculates the number of share reserves that are not reserved by open
+    /// positions.
+    fn calculate_idle_share_reserves(&self) -> FixedPoint {
+        let long_exposure = self.long_exposure().div_up(self.vault_share_price());
```
dpaiton, line 419: I'm confused about this -- `self.long_exposure()` is in units of shares, right? So the variable you're creating is `shares/shares` and is unitless? I think that's fine, it happens, but it's not `long_exposure` anymore right? If so I think the variable name should change to something else.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,164 @@ impl State {
...
+        let idle_shares = {
+            if self.share_reserves() > long_exposure + self.minimum_share_reserves() {
+                return self.share_reserves() - long_exposure - self.minimum_share_reserves();
+            }
+            fixed!(0)
```
dpaiton, line 424: I think it's more clear to return in each condition and not assign to a variable. e.g.:
```rs
if self.share_reserves() > long_exposure + self.minimum_share_reserves() {
  self.share_reserves() - long_exposure - self.minimum_share_reserves()
} else {
  fixed!(0)
}
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,164 @@ impl State {
...
+                self.bond_reserves() - bond_amount
+            } else {
+                return Err(eyre!("Calculating the bond reserves underflows"));
+            }
+        };
+
+        // NOTE: Round up since this is on the rhs of the final subtraction.
+        //
+        // derivative = c * (mu * z_e(x)) ** -t_s +
+        //              (y / z_e) * (y(x)) ** -t_s -
```
dpaiton, line 497: `z_e` should be `z_e(x)`?

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,164 @@ impl State {
...
+            if bond_amount < self.bond_reserves() {
+                self.bond_reserves() - bond_amount
+            } else {
+                return Err(eyre!("Calculating the bond reserves underflows"));
+            }
+        };
+
+        // NOTE: Round up since this is on the rhs of the final subtraction.
+        //
+        // derivative = c * (mu * z_e(x)) ** -t_s +
```
dpaiton, line 496: This uses `-t_s` but the code uses `self.time_stretch()`. In solidity we have to map `t_s` to `-t_s` but I think we do not in rust. Honestly I get confused by this so we may want to double check with @jalextowle , but either way I think the comment should match the code in Rust.

dpaiton: Also the code does `c / (...)` but the comment says `c * (...)`

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,164 @@ impl State {
...
+        );
+        if derivative < rhs {
+            return Err(eyre!("Derivative is less than right hand side"));
+        }
+        let derivative = derivative - rhs;
+
+        // NOTE: Round up since this is on the rhs of the final subtraction.
+        //
+        // inner = (
+        //             (mu / c) * (k(x) - (y(x) + dy) ** (1 - t_s))
```
dpaiton, line 521: this is not `k(x)` right? `k` is not dependent on `x`

crates/hyperdrive-math/src/lp.rs
```diff
@@ -353,6 +412,164 @@ impl State {
...
+    /// paased in because so that downstream function like kUp() can still be
+    /// used.
+    ///
+    /// Given a signed bond amount, this function calculates the negation
+    /// of the derivative of `calculateSharesOutGivenBondsIn` when the
+    /// bond amount is positive or the derivative of
+    /// `calculateSharesInGivenBondsOut` when the bond amount is negative.
+    /// In both cases, the calculation is given by:
+    ///
+    ///  derivative = (1 - zeta / z) * (
```
dpaiton, line 441: can you change `x` to `delta_y` (or `\Delta y` if in latex)? We've been somewhat inconsistent about this elsewhere, I'm trying to move us in the direction of consistently using `x` and `\Delta x` for "base".

dpaiton: assuming I have that right... seeing `y(x)` confuses me further.... Maybe the request is better stated as "can you make sure we use `z` for shares; `y` for bonds; and `x` for base)

crates/hyperdrive-math/src/lp.rs
sentilesdal: It must be in base and we're converting to shares.  This was translated from here:

https://github.com/delvtech/hyperdrive/blob/f76f5dacd38664cae6a2539770f30ff595d92141/contracts/src/internal/HyperdriveBase.sol#L575

crates/hyperdrive-math/src/lp.rs
sentilesdal: yup

crates/hyperdrive-math/src/lp.rs
sentilesdal: The comments are copied from the solidity.  I'd like to keep them the same, if there are things we'd like to change, then matching PRs should go up to fix both here and in solidity.

crates/hyperdrive-math/src/lp.rs
sentilesdal: it does match, the -1 part of the exponent is dropped and the base is flipped.  the code is altered from the comment to deal with lack of fixedpoints negative number capability.

crates/hyperdrive-math/src/lp.rs
sentilesdal: The comments are copied from the solidity.  I'd like to keep them the same, if there are things we'd like to change, then matching PRs should go up to fix both here and in solidity.

crates/hyperdrive-math/src/lp.rs
sentilesdal: same thing here, I wanna keep it exactly the same as the solidity to avoid drift.  love that you're looking at these though (and understanding haha).  can we keep these comment changes outta this PR and then we can sit down and update in both spots?  🙏

crates/hyperdrive-math/src/lp.rs
dpaiton: ahh ok, `z = x/c` so maybe just rename to `long_exposure_shares`?

crates/hyperdrive-math/src/lp.rs
dpaiton: yup sounds good. We already have issue for it:
https://github.com/delvtech/hyperdrive/issues/867
https://github.com/delvtech/hyperdrive/issues/965

can you link 965 in a TODO comment here?

crates/hyperdrive-math/src/lp.rs
dpaiton: Ahh so it's positive in the denominator; got it. Can you add a comment mentioning this? I'm trying to make sure that comments match code as much as possible, or have explanations whenever it doesn't.

crates/hyperdrive-math/src/lp.rs
dpaiton: roger; I commented in the issue mentioned above to look into this.

crates/hyperdrive-math/src/lp.rs
dpaiton: also fyi there's a general problem that the comment math doesn't always match what is implemented. It's residual from new people adding comments as part of their onboarding and doing it incorrectly :(

dpaiton (approved):
approved pending replies to our comments

crates/hyperdrive-math/src/lp.rs
sentilesdal: I have a PR up for this now, I'll update to use the new version when that PR lands.
https://github.com/delvtech/hyperdrive-rs/actions/runs/8897866606/job/24435373595?pr=11

crates/hyperdrive-math/src/lp.rs
sentilesdal: I think the name is right, marking as resolved.

PR #8, opened 2024-04-29
update latest tag script to be more general

dpaiton:
This sets a single version at the workspace level and uses a build script to prop it to lower crates. It also adds a script for CI that forces hyperdrive-rs to use the latest hyperdrive release when building.


slundqui (approved):

ryangoree (approved):
LGTM 🚀

PR #9, opened 2024-04-29
Adding python fuzz script that runs rust tests in a loop and logs to rollbar if it fails

slundqui:
# Resolved Issues
Using python, we run individual fuzz tests in a loop, and log any failed tests to rollbar.

# Review Checklists

- [ ] **Testing**


dpaiton:
This doesn't check all of the boxes in https://github.com/delvtech/hyperdrive/issues/1007 but I think it is enough to consider the task done. If we decide the other tasks have more utility we can revisit.

PR #11, opened 2024-04-30
Update `calculate_present_value` error handling.

sentilesdal:
# Description
Updating `calculate_present_value` with better error handling instead of throwing a panic.  We should convert the rest of the lower level functions to do the same.

I also included some Rust lint fixes that we were accumulating in the repo.

# Review Checklists

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -1,3 +1,5 @@
+use std::cmp::Ordering;
+
 use ethers::types::I256;
 use fixed_point::FixedPoint;
 use fixed_point_macros::fixed;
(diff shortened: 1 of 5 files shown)


crates/hyperdrive-math/src/short/max.rs
```diff
@@ -145,14 +147,14 @@ impl State {
...
                 self.short_deposit_derivative(max_bond_amount, spot_price, open_vault_share_price);
-            if deposit < target_budget {
-                max_bond_amount += (target_budget - deposit) / derivative;
-            } else if deposit > target_budget {
-                max_bond_amount -= (deposit - target_budget) / derivative;
-            } else {
-                // If we find the exact solution, we set the result and stop
-                best_valid_max_bond_amount = max_bond_amount;
-                break;
+            match deposit.cmp(&target_budget) {
```
ryangoree, line 150: TIL about `cmp`. neat

ryangoree (approved):
lgtm!

PR #13, opened 2024-04-30
Remove manual symlink/clone requirement

ryangoree:
Related to https://github.com/delvtech/hyperdrive/issues/978

# Description

This PR removes the requirement to manually symlink or clone the hyperdrive repo by modifying the `hyperdrive_wrappers` build script to clone it if it doesn't already exist using `HYPERDRIVE_REPO_URL` and `HYPERDRIVE_REF` environment variables.


crates/hyperdrive-wrappers/build.rs
```diff
@@ -56,40 +58,41 @@ const TARGETS: &[&str] = &[
...
-    }
-    Ok(artifacts)
-}
-
 fn main() -> Result<()> {
-    // Re-run this script whenever the build script itself or a contract changes.
+    // Re-run this script whenever .env, the build script itself, or a contract changes.
+    println!("cargo:rerun-if-changed=.env");
     println!("cargo:rerun-if-changed=build.rs");
     println!("cargo:rerun-if-changed=hyperdrive/contracts/");
```
dpaiton, line 65: `println!("cargo:rerun-if-changed=hyperdrive/contracts/");`
is this going to work if the repo is not cloned beforehand?

dpaiton: I think we can probably just remove this line

.github/workflows/rust_test.yml
```diff
@@ -38,16 +38,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: check out hyperdrive
```
dpaiton, line 41: if this works then you can remove the code above
```
      - name: Fetch latest release tag from Hyperdrive
        run: ./scripts/fetch-latest-tag.sh delvtech hyperdrive
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/rust_lint.yml
```diff
@@ -38,16 +38,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: check out hyperdrive
```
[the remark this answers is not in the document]
dpaiton: And I guess remove the script, although I was proud of that script :(

crates/hyperdrive-wrappers/.env.example
```diff
@@ -0,0 +1,2 @@
+HYPERDRIVE_REPO_URL=https://github.com/delvtech/hyperdrive.git
```
dpaiton, line 1: is there going to be an install step to move this to `.env` or do we read directly from `.env.example`? If the latter, why not just call it `.env.`?

dpaiton (approved):
awesome, much cleaner!

You probably want to update the readme. We still need devs to symlink, so maybe readme is fine. But it could be worth noting that CI handles this by pulling the env version directly.

.github/workflows/rust_test.yml
ryangoree: This is still needed to set the `HYPERDRIVE_REF` environment variable to the latest release below.

.github/workflows/rust_lint.yml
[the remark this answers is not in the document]
ryangoree: We still need it actually! See earlier comment. And it *is* a cool script 💪

ryangoree:
> awesome, much cleaner!
> 
> You probably want to update the readme. We still need devs to symlink, so maybe readme is fine. But it could be worth noting that CI handles this by pulling the env version directly.

🙌 

I updated the README to include instructions on setting up your environment and moved the symlinking instructions to a callout (quote) about local development.

PR #15, opened 2024-05-01
fix `calculate_spot_price_after_short` and add open short deltas & preview

dpaiton:
# Resolved Issues
No issue

# Description
`calculate_spot_price_after_short` did not correctly compute the share deltas when a `base_amount` was not given. This PR adds a test to demonstrate this as well as a fix.

Part of the process also required adding `calculate_open_short_deltas` and a corresponding test.

Since I was most of the way there, I also added a `preview_open_short` that returns a mutated state after the reserve deltas have been applied.

# Review Checklists

- [ ] **Testing**

### DIFF
--- a/bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
+++ b/bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
@@ -150,11 +150,11 @@ impl HyperdriveState {
 
     pub fn calculate_open_short(
         &self,
-        short_amount: &str,
+        bond_amount: &str,
         open_vault_share_price: &str,
     ) -> PyResult<String> {
-        let short_amount_fp = FixedPoint::from(U256::from_dec_str(short_amount).map_err(|_| {
-            PyErr::new::<PyValueError, _>("Failed to convert short_amount string to U256")
+        let bond_amount_fp = FixedPoint::from(U256::from_dec_str(bond_amount).map_err(|_| {
+            PyErr::new::<PyValueError, _>("Failed to convert bond_amount string to U256")
         })?);
         let open_vault_share_price_fp =
             FixedPoint::from(U256::from_dec_str(open_vault_share_price).map_err(|_| {
@@ -164,19 +164,19 @@ impl HyperdriveState {
             })?);
         let result_fp = self
             .state
-            .calculate_open_short(short_amount_fp, open_vault_share_price_fp)
+            .calculate_open_short(bond_amount_fp, open_vault_share_price_fp)
             .unwrap();
         let result = U256::from(result_fp).to_string();
         Ok(result)
     }
 
-    pub fn calculate_open_short_share_reserves_delta(&self, bond_amount: &str) -> PyResult<String> {
+    pub fn calculate_pool_deltas_after_open_short(&self, bond_amount: &str) -> PyResult<String> {
         let bond_amount_fp = FixedPoint::from(U256::from_dec_str(bond_amount).map_err(|_| {
             PyErr::new::<PyValueError, _>("Failed to convert bond_amount string to U256")
         })?);
         let result_fp = self
             .state
-            .calculate_open_short_share_reserves_delta(bond_amount_fp)
+            .calculate_pool_deltas_after_open_short(bond_amount_fp)
             .unwrap();
         let result = U256::from(result_fp).to_string();
         Ok(result)
--- a/bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
+++ b/bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
@@ -71,9 +71,10 @@ impl HyperdriveState {
         let result_fp = self
             .state
             .calculate_spot_price_after_short(bond_amount_fp, maybe_base_amount_fp)
-            .unwrap();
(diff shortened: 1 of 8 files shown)


bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
```diff
@@ -169,6 +170,18 @@ impl HyperdriveState {
...
     }
 
+    pub fn calculate_open_short_deltas(&self, bond_amount: &str) -> PyResult<String> {
+        let bond_amount_fp = FixedPoint::from(U256::from_dec_str(bond_amount).map_err(|_| {
+            PyErr::new::<PyValueError, _>("Failed to convert bond_amount string to U256")
+        })?);
+        let result_fp = self
+            .state
+            .calculate_open_short_deltas(bond_amount_fp)
+            .unwrap();
```
sentilesdal, line 180: don't we want to handle the error here with a `map_err()`?

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -314,6 +336,70 @@ mod tests {
...
+                    state.vault_share_price().into(),
+                    state.initial_vault_share_price().into(),
+                )
+                .call()
+                .await
+            {
+                Ok(expected) => {
+                    let expected_fp = FixedPoint::from(expected);
+                    // short principal is too low to account for fees
+                    if expected_fp < fees {
```
sentilesdal, line 389: nit use a match compare here to avoid rust lint warnings:

```rust
match expected_fp.cmp(fees) {
    Ordering::Less => assert!(actual.is_err()),
    _ => assert_eq!(actual.unwrap(), FixedPoint::from(expected) - fees),
}

File: crates/hyperdrive-math/src/short/open.rs [hunk 3]
```diff
@@ -548,6 +634,63 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn fuzz_calculate_spot_price_after_short_defaults() -> Result<()> {
```
sentilesdal, line 638: after_short_deltas?

sentilesdal (approved):
overall looks good!  I think we should get in the habit of adding /// comments at the top of functions to give clues what they do, helps reviewers at least.

crates/hyperdrive-math/src/short/open.rs
dpaiton: no it's a specific test for the default behavior. But I'll move the word to make it more clear.

crates/hyperdrive-math/src/short/open.rs
dpaiton: I'm gonna add a global ignore for this one bc I think the original syntax is more readable.

bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
```diff
@@ -169,6 +170,18 @@ impl HyperdriveState {
         Ok(result)
     }
 
+    pub fn calculate_open_short_deltas(&self, bond_amount: &str) -> PyResult<String> {
+        let bond_amount_fp = FixedPoint::from(U256::from_dec_str(bond_amount).map_err(|_| {
+            PyErr::new::<PyValueError, _>("Failed to convert bond_amount string to U256")
```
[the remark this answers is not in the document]
dpaiton: no, we map error on the U256 from. I guess we could also put a map error on the FixedPoint from, but that's not how we've done it so far.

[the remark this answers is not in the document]
dpaiton: ultimately yeah the thing we're trying to do is convert to fixedpoint and it broke, so I agree it makes sense to have that in the error. But the way we do it everywhere is assume that the error will come from the string->U256 conversion and not the U256->FixedPoint conversion. Since we have the error attached to the string->U256 conversion we say that in the error message.

Not gonna make the change here to keep things consistent, but yeah I think we could do this better in a single PR.

bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
dpaiton: yeah I like that a lot better. The best example is in `calculate_targeted_long_with_budget`, where we do:

```
            .map_err(|err| {
                PyErr::new::<PyValueError, _>(format!(
                    "Calculate_targeted_long_with_budget returned the error: {:?}",
                    err
                ))
            })?;
```

But that's the only place we do it. Again I think it'd be better to fix this everyone in a single PR. I'll make note & do it.

https://github.com/delvtech/hyperdrive-rs/issues/64

bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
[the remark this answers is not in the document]
dpaiton: https://github.com/delvtech/hyperdrive-rs/issues/65

crates/hyperdrive-math/src/short/open.rs
dpaiton: https://github.com/delvtech/hyperdrive-rs/issues/66

PR #17, opened 2024-05-01
Add `distribute_excess_idle`

sentilesdal:
# Resolved Issues

# Description
Adds `distribute_excess_idle` and all the helpers needed for it.  Here's the full list:

- `calculate_distribute_excess_idle`
- `calculate_distribute_excess_idle_share_proceeds`
- `calculate_distribute_express_idle_withdrawal_shares_redeemed`
- `should_short_circuit_distribute_excess_idle_share_proceeds`
- `calculate_distribute_excess_idle_share_proceeds_net_long_edge_case`
- `calculate_max_share_reserves_delta`

explicit tests were added:
- `fuzz_test_calculate_distribute_excess_idle`
- `fuzz_test_calculate_distribute_excess_idle_withdrawal_shares_redeemed`
- ` fuzz_test_should_short_circuit_distribute_excess_idle_share_proceeds`
- `fuzz_test_calculate_distribute_excess_idle_share_proceeds_net_long_edge_case`

Everything should be covered by the first test, but the others helped me debug along the way.

# Review Checklists

- [ ] **Testing**


crates/hyperdrive-math/src/lp.rs
```diff
@@ -207,11 +212,7 @@ impl State {
         }
 
         // Get the updated share reserves.
-        let new_share_reserves = if share_reserves_delta > I256::zero() {
-            I256::try_from(share_reserves).unwrap() + share_reserves_delta
-        } else {
-            I256::try_from(share_reserves).unwrap() - share_reserves_delta
-        };
+        let new_share_reserves = I256::try_from(share_reserves)? + share_reserves_delta;
```
dpaiton, line 215: good catch!

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+        current_block_timestamp: U256,
+        active_lp_total_supply: FixedPoint,
+        withdrawal_shares_total_supply: FixedPoint, // withdraw shares - ready to withdraw
+        max_iterations: u64,
+    ) -> Result<(FixedPoint, FixedPoint)> {
+        // Steps 1 and 2: Calculate the maximum amount the share reserves can be
+        // debited. If the effective share reserves or the maximum share
+        // reserves delta can't be calculated or if the maximum share reserves
+        // delta is zero, idle can't be distributed.
+        // doesn't need original effective share reserves, gets it from state
```
dpaiton, line 699: >         // doesn't need original effective share reserves, gets it from state
not sure what this is talking about?

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+    /// reserves and dl is the amount of LP shares to be burned, we can
+    /// derive the calculation as follows:
+    ///
+    ///  PV(0) / l = PV(dx) / (l - dl)
+    ///                =>
+    ///  dl = l - l * (PV(dx) / PV(0))
+    ///
+    ///  We round this calculation up to err on the side of slightly too
+    ///  many withdrawal shares being redeemed.
+    fn calculate_distribute_express_idle_withdrawal_shares_redeemed(
```
dpaiton, line 1058: is that supposed to be "express_idle" and not "excess_idle"?

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+            // we're done and we pay out nothing.
+            if withdrawal_shares_redeemed == fixed!(0) {
+                return Err(eyre!(
+                    "Withdrawal shares redeemed from excess idle is zero."
+                ));
+
+            // Step 3: Otherwise if this amount is less than or equal to the amount
+            // of withdrawal shares outstanding, then we're done and we pay out the
+            // full maximum share reserves delta.
+            } else if withdrawal_shares_redeemed <= withdrawal_shares_total_supply {
```
dpaiton, line 730: preferred formatting on is to separate the closing bracket from the "else if":

```rs
if thing {
  ...
}

// comment
else if other_thing {
  ...
}

// comment
else {
  ...
}
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+        current_block_timestamp: U256,
+        active_lp_total_supply: FixedPoint, // just the number of lp tokens
+        withdrawal_shares_total_supply: FixedPoint,
+        max_share_reserves_delta: FixedPoint,
+        max_iterations: u64,
+    ) -> Result<FixedPoint> {
+        let net_curve_trade =
+            self.calculate_net_curve_trade_from_timestamp(current_block_timestamp)?;
+        // Calculate the LP total supply.
+        let lp_total_supply = active_lp_total_supply + withdrawal_shares_total_supply;
```
dpaiton, line 780: above we define `lp_total_supply = active_lp_total_supply + withdrawal_shares_total_supply - withdrawal_shares_ready_to_withdraw`. Do we know that `withdrawal_shares_ready_to_withdraw=0` here? If so, can you comment it so we are consistent in our definitions?

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+        let starting_present_value = self.calculate_present_value(current_block_timestamp)?;
+
+        // NOTE: Round the initial guess down to avoid overshooting.
+        //
+        // We make an initial guess for Newton's method by assuming that the
+        // ratio of the share reserves delta to the withdrawal shares
```
dpaiton, line 786: this comment drops off, what are we assuming?

dpaiton: oh maybe it's just that this NOTE is not supposed to be here?

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+        let mut closest_share_proceeds = fixed!(0);
+        let mut closest_present_value = fixed!(0);
+        for _ in 0..max(max_iterations, SHARE_PROCEEDS_MAX_ITERATIONS) {
+            // Clamp the share proceeds to the max share reserves delta since
+            // values above this threshold are always invalid.
+            share_proceeds = min(share_proceeds, max_share_reserves_delta);
+
+            // Simulate applying the share proceeds to the reserves.
+            //
+            // NOTE: If the updated reserves can't be calculated,  we can't
```
dpaiton, line 820: extra space after comma

dpaiton: also I'm not sure what this note is adding. Are you saying that newton's method cannot be performed if we can't calculate the updated reserves? Why would this happen?

dpaiton: I see a similar note below on present value. I guess they're fine to leave but it seems unnecessary. Any time we use a `?` we're saying "if this can't be done, the loop breaks", no need to include a comment to that effect.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+                return Ok(share_proceeds);
+            }
+
+            // If the pool is net long, we can solve for the next iteration of
+            // Newton's method directly when the net curve trade is greater than
+            // or equal to the max bond amount.
+            if net_curve_trade > I256::zero() {
+                // Calculate the max bond amount.
+                // NOTE: If the max bond amount couldn't be calculated, we
+                // can't continue the calculation. Return 0 to indicate that
```
dpaiton, line 849: this note seems incorrect; we're not returning 0 we're returning an error

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+                smallest_delta = delta;
+                closest_share_proceeds = share_proceeds;
+                closest_present_value = present_value;
+            }
+
+            // We calculate the updated share proceeds `x_n+1` by proceeding
+            // with Newton's method. This is given by:
+            //
+            // x_n+1 = x_n - F(x_n) / F'(x_n)
+            match delta.cmp(&I256::zero()) {
```
dpaiton, line 922: I haven't had a chance to get around to this, but we don't like this syntax and will be ignoring the lint error. I don't think it is required to pass CI anyway. If possible, please change to the usual syntax and if not please add a TODO to change it back so we can find it later.

https://github.com/delvtech/hyperdrive-rs/issues/66

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+    }
+
+    /// Calculates the amount of withdrawal shares that can be redeemed and
+    /// the share proceeds the withdrawal pool should receive given the
+    /// pool's current idle liquidity. We use the following algorithm to
+    /// ensure that the withdrawal pool receives the correct amount of
+    /// shares to (1) preserve the LP share price and (2) pay out as much
+    /// of the idle liquidity as possible to the withdrawal pool:
+    ///
+    /// 1. If `y_s * t_s <= y_l * t_l` or
```
dpaiton, line 675: For function-level docstrings (prefaced with `///`) we use latex notation and `$$` to define equations, as well as "`\cdot`" instead of "`*`". For example,

```
/// 1. if $y_s \cdot t_s <= y_l \cot t_l$ or
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+        }
+    }
+
+    /// Calculates the amount of withdrawal shares that can be redeemed
+    /// given an amount of shares to remove from the share reserves.
+    /// Assuming that dz is the amount of shares to remove from the
+    /// reserves and dl is the amount of LP shares to be burned, we can
+    /// derive the calculation as follows:
+    ///
+    ///  PV(0) / l = PV(dx) / (l - dl)
```
dpaiton, line 1052: please use latex and surround in $$, e.g.:

```
///
/// $$
/// \frac{PV(0)}{l} = \frac{PV(\Delta x)}{l - \Delta l}
/// $$
///
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+            }
+            return Ok(max_share_reserves_delta);
+        } else {
+            return Err(eyre!("max_scaling_factor greater than 1."));
+        }
+    }
+
+    /// Calculates the amount of withdrawal shares that can be redeemed
+    /// given an amount of shares to remove from the share reserves.
+    /// Assuming that dz is the amount of shares to remove from the
```
dpaiton, line 1048: this says `dz` but you use `dx` in the math below.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+        //
+        // Calculate the amount of withdrawal shares that can be redeemed.
+        let lp_total_supply = active_lp_total_supply + withdrawal_shares_total_supply;
+        Ok(lp_total_supply
+            - lp_total_supply.mul_div_down(ending_present_value, starting_present_value))
+    }
+
+    /// One of the edge cases that occurs when using Newton's method for the
+    /// share proceeds while distributing excess idle is when the net curve
+    /// trade is larger than the max bond amount. In this case, the
```
dpaiton, line 1093: comment ends mid sentence

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+        // calculate the share proceeds. This should never happen, but for
+        // safety we return a failure flag and break the loop at this point.
+        if original_share_adjustment <= I256::zero() {
+            return Err(eyre!("Original share adjustment is zero or negative."));
+        }
+        // Calculate the net flat trade.
+        let net_flat_trade = self.calculate_net_flat_trade(
+            self.long_average_maturity_time(),
+            self.short_average_maturity_time(),
+            // NOTE: Round the exponent up since this rounds the result up.
```
dpaiton, line 1112: spurious note?

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,540 @@ impl State {
...
+            if net_flat_trade >= I256::zero() {
+                if net_flat_trade < rhs.try_into()? {
+                    rhs - net_flat_trade.into()
+                } else {
+                    return Err(eyre!("Underflow in calculating rhs."));
+                }
+            } else {
+                rhs + (-net_flat_trade).into()
+            }
+            // NOTE: Round the exponent up since this rounds the result up.
```
dpaiton, line 1140: what is this note referring to?

crates/hyperdrive-math/src/lp.rs
sentilesdal: express idle sounds better 😂

crates/hyperdrive-math/src/lp.rs
dpaiton: e.g. here: https://github.com/delvtech/hyperdrive-rs/blob/main/crates/hyperdrive-math/src/long/targeted.rs#L81

crates/hyperdrive-math/src/lp.rs
sentilesdal: That was a note to myself, removed it.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -1047,9 +1603,239 @@ mod tests {
...
+            // We clip withdrawal_shares_total_supply to ensure active_lp_total_supply doesn't underflow.
+            let active_lp_total_supply = present_state.lp_total_supply()
+                + present_state.withdrawal_shares_ready_to_withdraw();
+            let withdrawal_shares_total_supply = rng.gen_range(fixed!(0)..=active_lp_total_supply);
+            let active_lp_total_supply = active_lp_total_supply - withdrawal_shares_total_supply;
+            // This errors out a lot so we need to catch that here.
+            let starting_present_value =
+                match present_state.calculate_present_value(U256::from(current_block_timestamp)) {
+                    Ok(result) => result,
+                    Err(_) => continue,
```
dpaiton, line 1637: one thing I've done in the past is keep track of a counter of fuzz runs and assert that it is > 0 at the end of the `for` loop. This is probably overkill, but technically `calculate_present_value` could fail every time and the test would pass. 

e.g. here: https://github.com/delvtech/hyperdrive-rs/blob/main/crates/hyperdrive-math/src/short/open.rs#L653-L703

crates/hyperdrive-math/src/lp.rs
sentilesdal: yeah, that NOTE snuck in

sentilesdal: removed it

crates/hyperdrive-math/src/lp.rs
sentilesdal: yeah i agree, i'll remove it

crates/hyperdrive-math/src/lp.rs
sentilesdal: Added the full comment from solidity

crates/hyperdrive-math/src/lp.rs
sentilesdal: ya

crates/hyperdrive-math/src/lp.rs
sentilesdal: removed

crates/hyperdrive-math/src/lp.rs
```diff
@@ -1149,4 +1946,162 @@ mod tests {
...
+
+        // Fuzz the rust and solidity implementations against each other.
+        let mut rng = thread_rng();
+
+        for _ in 0..*FAST_FUZZ_RUNS {
+            // Generate random states.
+            let original_state = rng.gen::<State>();
+            let mut present_state = rng.gen::<State>();
+
+            //**************Get Variables*****************//
```
dpaiton, line 2020: this is new 👀 we don't use this type of comment to differentiate logic blocks anywhere else, probably best not to start here.

crates/hyperdrive-math/src/lp.rs
sentilesdal: Yeah I do that too while writing the tests to make sure that I'm seeing success cases.  I dunno that we want it in  everywhere though because it might cause flakes

crates/hyperdrive-math/src/lp.rs
sentilesdal: @jalextowle In solidity we define `lpTotalSupply` as
```solidity
        uint256 lpTotalSupply = _params.activeLpTotalSupply +
            _params.withdrawalSharesTotalSupply;
  ```
  in `calculateDistributeExcessIdleWithdrawalSharesRedeemed()` and `calculateDistributeExcessIdleShareProceeds()`
  
  and as 
  ```solidity
          uint256 lpTotalSupply = _totalSupply[AssetId._LP_ASSET_ID] +
            _totalSupply[AssetId._WITHDRAWAL_SHARE_ASSET_ID] -
            _withdrawPool.readyToWithdraw;
  ```
in `getPoolConfig()` and `_addLiquidity()`

Just double checking that this is intentional.

Comment (sentilesdal) [reply crates/hyperdrive-math/src/lp.rs:780]: @dpaiton I'll rename these internal ones to something else, so far I've got `lp_total_supply_and_withdrawal_shares_ready_to_withdraw` but I'm willing to workshop it 😂

Comment (dpaiton) [reply crates/hyperdrive-math/src/lp.rs:780]: lol `let active_lp_total_supply_plus_withdrawal_shares_total_supply = active_lp_total_supply + withdrawal_shares_total_supply` should be sufficiently descriptive 😆

File: crates/hyperdrive-math/src/lp.rs (hunk 16 above)
Comment (sentilesdal) [reply crates/hyperdrive-math/src/lp.rs:1664]: I just checked and most of these are hitting success over 1/2 the time so im gonna leave it out

File: crates/hyperdrive-math/src/lp.rs [hunk 18]
```diff
@@ -600,6 +621,567 @@ impl State {
...
+        withdrawal_shares_total_supply: FixedPoint, // withdraw shares - ready to withdraw
+        max_iterations: u64,
+    ) -> Result<(FixedPoint, FixedPoint)> {
+        // Steps 1 and 2: Calculate the maximum amount the share reserves can be
+        // debited. If the effective share reserves or the maximum share
+        // reserves delta can't be calculated or if the maximum share reserves
+        // delta is zero, idle can't be distributed.
+        let max_share_reserves_delta =
+            self.calculate_max_share_reserves_delta(current_block_timestamp)?;
+
```
jalextowle, line 701: Nit: Remove

```suggestion
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+        // maximum share reserves delta that was previously calculated, then
+        // we can't distribute excess idle since we ruled out the possibility
+        // of paying out the full maximum share reserves delta in step 3.
+        if share_proceeds >= max_share_reserves_delta {
+            return Err(eyre!(
+                "Share proceeds greater than or equal to the maximum share reserves delta."
+            ));
+        }
+
+        Ok((withdrawal_shares_redeemed, share_proceeds))
```
jalextowle, line 765: Nit: We're missing a comment here for Solidity parity:

```suggestion
        // Step 5: Return the amount of withdrawal shares redeemed and the
        // share proceeds.
        Ok((withdrawal_shares_redeemed, share_proceeds))
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+    }
+
+    /// Calculates the share proceeds to distribute to the withdrawal pool
+    /// assuming that all of the outstanding withdrawal shares will be
+    /// redeemed. The share proceeds are calculated such that the LP share
+    /// price is conserved. When we need to round, we round down to err on
+    /// the side of slightly too few shares being paid out.
```
jalextowle, line 772: I think the math is helpful. Let's add this back into the comment:

```
    ///      PV(0) / l = PV(dx) / (l - dl)
    ///                =>
    ///      dl = l - l * (PV(dx) / PV(0))
```

My default is assumption is that the Rust and Solidity should match

jalextowle: Nvm, I'm looking at the wrong function. IMO, this would be clearer/easier to review if we maintain the same order of functions as we had in Solidity. This should be `calculate_distribute_excess_idle_withdrawal_shares_redeemed`

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+            // Clamp the share proceeds to the max share reserves delta since
+            // values above this threshold are always invalid.
+            share_proceeds = min(share_proceeds, max_share_reserves_delta);
+
+            // Simulate applying the share proceeds to the reserves.
+            //
+            // NOTE: We are calling this with 'self' so that original values are
+            // used with the updated value of share_proceeds
+            let updated_state =
+                self.get_state_after_liquidity_update(-I256::try_from(share_proceeds)?)?;
```
jalextowle, line 826: If we want to maintain parity with Solidity, it might be better to `unwrap` instead of bubbling up the `I256::try_from` error. This is just a note since I actually like the way you do it here more, but in Solidity, the entire calculation would be halted.

jalextowle: Another way you could do this is to return a different error that indicates to the caller (`calculate_distribute_excess_idle`) that we hit an unrecoverable error. Otherwise, this could make it look like the calculation will succeed when it actually fails.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+    ///    using Newton's method.
+    /// 3. Set `dw = (1 - PV(dz_max) / PV(0)) * l`. If `dw <= w`, then
+    ///    proceed to step (5). Otherwise, set `dw = w` and continue to
+    ///    step (4).
+    /// 4. Solve `PV(0) / l = PV(dz) / (l - dw)` for `dz` using Newton's
+    ///    method if `y_l * t_l != y_s * t_s` or directly otherwise.
+    /// 5. Return `dw` and `dz`.
+    ///
+    ///  Returns (withdrawal_shares_redeemed, share_proceeds)
+    pub fn calculate_distribute_excess_idle(
```
jalextowle, line 688: This is meta-comment that applies to a bunch of lines. I'm happy to clarify which ones I'm talking about if you aren't sure about some cases. 

In Solidity, we use the pattern:

```
(uint256 X, bool success) = calculateBlahBlahBlahSafe(someArgs);
if (!success) {
    return (0, false);
}
```

in a bunch of cases. This is roughly equivalent to the `?` operator in Rust:

```
let X = calculate_blah_blah_blah_safe(some_args)?;
```

Where this breaks down is in `calculate_distribute_excess_idle`. In Solidity, `calculate_distribute_excess_idle` is NOT a safe function, so it doesn't return a failure flag. The implication of this is that `calculate_distribute_excess_idle` doesn't just unwrap all of the underlying results (this is an analogy, but stick with me for a sec) using `?`. Instead, it returns `0`  when a failure flag is returned. Here's an example of that: https://github.com/delvtech/hyperdrive/blob/a98362d1ead7c068957d1bdcfed9c6d7b0c7e4e8/contracts/src/libraries/LPMath.sol#L587.

With this in mind, there are a bunch of errors that this implementation will treat as failures that Solidity would handle as "no idle can be distributed."

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+    ) -> Result<(FixedPoint, FixedPoint)> {
+        // Steps 1 and 2: Calculate the maximum amount the share reserves can be
+        // debited. If the effective share reserves or the maximum share
+        // reserves delta can't be calculated or if the maximum share reserves
+        // delta is zero, idle can't be distributed.
+        let max_share_reserves_delta =
+            self.calculate_max_share_reserves_delta(current_block_timestamp)?;
+
+        if max_share_reserves_delta == fixed!(0) {
+            return Err(eyre!("Max share reserves delta is zero."));
```
jalextowle, line 703: We should just return `0` here.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+        // anything.
+        let share_proceeds = self.calculate_distribute_excess_idle_share_proceeds(
+            current_block_timestamp,
+            active_lp_total_supply,
+            withdrawal_shares_total_supply,
+            max_share_reserves_delta,
+            max_iterations,
+        )?;
+        if share_proceeds == fixed!(0) {
+            return Err(eyre!("Share proceeds from excess idle are zero."));
```
jalextowle, line 752: Same here, we should return `(0, 0)`. I'll comment this where I see it, but it would be good to do a scan through the PR since I might miss a case where this happens.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+        }
+
+        // Step 4: If the share proceeds are greater than or equal to the
+        // maximum share reserves delta that was previously calculated, then
+        // we can't distribute excess idle since we ruled out the possibility
+        // of paying out the full maximum share reserves delta in step 3.
+        if share_proceeds >= max_share_reserves_delta {
+            return Err(eyre!(
+                "Share proceeds greater than or equal to the maximum share reserves delta."
+            ));
```
jalextowle, line 762: We should return `(0, 0)` here.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+                present_value,
+            )? {
+                return Ok(share_proceeds);
+            }
+
+            // If the pool is net long, we can solve for the next iteration of
+            // Newton's method directly when the net curve trade is greater than
+            // or equal to the max bond amount.
+            if net_curve_trade > I256::zero() {
+                // Calculate the max bond amount.
```
jalextowle, line 845: The comment is Solidity is:

```
                // Calculate the max bond amount. If the calculation fails, we
                // return a failure flag.
```

I get that we do this implicitly using `?`, but this context is still helpful IMO

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+                let max_bond_amount = updated_state
+                    .calculate_max_sell_bonds_in_safe(self.minimum_share_reserves())?;
+
+                // If the net curve trade is greater than or equal to the max
+                // bond amount, we can solve directly for the share proceeds.
+                let net_curve_trade = FixedPoint::from(U256::try_from(net_curve_trade)?);
+                if net_curve_trade >= max_bond_amount {
+                    let share_proceeds = updated_state
```
jalextowle, line 853: We're missing the corresponding Solidity comment here:

```
                    // Solve the objective function directly assuming that it is
                    // linear with respect to the share proceeds.
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+                        .calculate_distribute_excess_idle_share_proceeds_net_long_edge_case(
+                            self.share_adjustment(),
+                            self.share_reserves(),
+                            starting_present_value,
+                            active_lp_total_supply,
+                            withdrawal_shares_total_supply,
+                        )?;
```
jalextowle, line 860: In Solidity, we do this:

```
                    if (!success) {
                        // NOTE: Return 0 to indicate that the share proceeds
                        // couldn't be calculated.
                        return 0;
                    }
```

I think it would be good to match this as closely as possible. This would help in teasing out which errors are unrecoverable and which aren't at the level of `calculate_distribute_excess_idle`.

crates/hyperdrive-math/src/lp.rs
jalextowle: > @jalextowle In solidity we define `lpTotalSupply` as
> 
> ```solidity
>         uint256 lpTotalSupply = _params.activeLpTotalSupply +
>             _params.withdrawalSharesTotalSupply;
> ```
> 
> in `calculateDistributeExcessIdleWithdrawalSharesRedeemed()` and `calculateDistributeExcessIdleShareProceeds()`
> 
> and as
> 
> ```solidity
>         uint256 lpTotalSupply = _totalSupply[AssetId._LP_ASSET_ID] +
>           _totalSupply[AssetId._WITHDRAWAL_SHARE_ASSET_ID] -
>           _withdrawPool.readyToWithdraw;
> ```
> 
> in `getPoolConfig()` and `_addLiquidity()`
> 
> Just double checking that this is intentional.

Yeah this is intentional. We precompute this before `calculate_distribute_excess_idle`:

```
          _totalSupply[AssetId._WITHDRAWAL_SHARE_ASSET_ID] -
          _withdrawPool.readyToWithdraw;
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+        // invariant after all of the withdrawal shares are redeemed. If the
+        // calculation returns a share proceeds of zero, we can't pay out
+        // anything.
+        let share_proceeds = self.calculate_distribute_excess_idle_share_proceeds(
+            current_block_timestamp,
+            active_lp_total_supply,
+            withdrawal_shares_total_supply,
+            max_share_reserves_delta,
+            max_iterations,
+        )?;
```
jalextowle, line 750: Double-checking this, this `?` looks correct since `calculate_distribute_excess_idle_share_proceeds` isn't "safe" in Solidity: https://github.com/delvtech/hyperdrive/blob/main/contracts/src/libraries/LPMath.sol#L624. With this in mind, we'll need to be more careful about the values that we return inside of this function call. Most of the time, it returns zero when an underlying "safe" function fails, but there are a few places where returning an `Err` is appropriate (like with `I256::try_from`).

crates/hyperdrive-math/src/lp.rs
jalextowle: Now that I'm deeper into the review, I like doing this with `?`. I think we just need to be more careful with some of the other functions that return `Result`.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+            // NOTE: Round the LHS down to make the check stricter.
+            new_lp_share_price_down >= old_lp_share_price
+            // Ensure that new LP share price is less than or equal to the
+            // previous LP share price plus the minimum tolerance:
+            //
+            // PV_1 / l_1 <= (PV_0 / l_0) * (1 + tolerance)
+            //
+            // NOTE: Round the LHS up to make the check stricter.
+            && new_lp_share_price_up <= old_lp_share_price_with_tolerance,
+        )
```
jalextowle, line 665: It would be better if this more closely matched the Solidity IMO. Re-ordering the logic makes this quite a bit harder to review.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+
+                    // If the max bond amount is less than or equal to the net
+                    // curve trade, then Newton's method has terminated since
+                    // proceeding to the next step would result in reaching the
+                    // same point.
+                    if max_bond_amount <= net_curve_trade {
+                        return Ok(share_proceeds);
+                    }
+                    // Otherwise, we continue to the next iteration of Newton's
+                    // method.
```
jalextowle, line 880: I think we need a `continue` here. We want to skip the derivative part of the calculation and go to the next iteration of the for-loop.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+            // is net short.
+            let derivative = updated_state.calculate_shares_delta_given_bonds_delta_derivative(
+                net_curve_trade,
+                self.share_reserves(),
+                self.bond_reserves(),
+                self.effective_share_reserves(),
+                self.share_adjustment(),
+            )?;
+            if derivative >= fixed!(1e18) {
+                return Err(eyre!("Derivative is greater than one"));
```
jalextowle, line 896: Return `0` here and return `0` if the above function errors out.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+                        .div_down(derivative)
+                        .div_down(lp_total_supply);
+            } else if delta < I256::zero() {
+                let delta = FixedPoint::from(-delta)
+                    .div_down(derivative)
+                    .div_down(lp_total_supply);
+                if delta < share_proceeds {
+                    share_proceeds = share_proceeds - delta;
+                } else {
+                    return Err(eyre!("Delta is greater than share proceeds."));
```
jalextowle, line 931: Return `0` here.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+                }
+            } else {
+                break;
+            }
+        }
+
+        let updated_state =
+            self.get_state_after_liquidity_update(-I256::try_from(share_proceeds)?)?;
+        let present_value = updated_state.calculate_present_value(current_block_timestamp)?;
```
jalextowle, line 940: Looks like we're missing a comment here.

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+        if closest_present_value.div_down(active_lp_total_supply)
+            < starting_present_value.mul_div_up(
+                fixed!(1e18) - FixedPoint::from(SHARE_PROCEEDS_TOLERANCE),
+                lp_total_supply,
+            )
+            || closest_present_value.div_up(active_lp_total_supply)
+                > starting_present_value.mul_div_down(
+                    fixed!(1e18) + FixedPoint::from(SHARE_PROCEEDS_TOLERANCE),
+                    lp_total_supply,
+                )
```
jalextowle, line 963: This is what the solidity looks like:

```
        if (
            // NOTE: Round down to make the check stricter.
            closestPresentValue.divDown(params.activeLpTotalSupply) <
            params.startingPresentValue.mulDivUp(
                ONE - SHARE_PROCEEDS_TOLERANCE,
                lpTotalSupply
            ) ||
            // NOTE: Round up to make the check stricter.
            closestPresentValue.divUp(params.activeLpTotalSupply) >
            params.startingPresentValue.mulDivDown(
                ONE + SHARE_PROCEEDS_TOLERANCE,
                lpTotalSupply
            )
        ) {
            // NOTE: Return 0 to indicate that the share proceeds couldn't be
            // calculated.
            return 0;
        }
```

crates/hyperdrive-math/src/lp.rs
```diff
@@ -600,6 +621,567 @@ impl State {
...
+        // calculate the share proceeds. This should never happen, but for
+        // safety we return a failure flag and break the loop at this point.
+        if original_share_adjustment <= I256::zero() {
+            return Err(eyre!("Original share adjustment is zero or negative."));
+        }
+        // Calculate the net flat trade.
+        let net_flat_trade = self.calculate_net_flat_trade(
+            self.long_average_maturity_time(),
+            self.short_average_maturity_time(),
+        )?;
```
jalextowle, line 1141: Shouldn't this be `time_remaining` instead of `maturity_time`?

sentilesdal: yes! good catch.  I guess this passed because the timestamp was always zero in the tests.

crates/hyperdrive-math/src/lp.rs
sentilesdal: Ok, made the rust match this as close as prettier lets me

crates/hyperdrive-math/src/lp.rs
sentilesdal: I reorganzied the functions to match the order of LPMath.sol

crates/hyperdrive-math/src/lp.rs
sentilesdal: Ok I'm resolving this one, sounds like there's nothing to do there.

crates/hyperdrive-math/src/lp.rs
sentilesdal: Ok I went through the PR and it much more closely aligns with the solidity now!

crates/hyperdrive-math/src/lp.rs
sentilesdal: I still need to do this.

dpaiton (approved):

crates/hyperdrive-math/src/lp/math.rs
```diff
@@ -281,6 +301,652 @@ impl State {
...
+        let success = match self.effective_share_reserves_safe() {
+            Ok(_) => true,
+            // The error is safe from the calculation, panics are not.
+            Err(_) => false,
+        };
+        if success == false {
+            return Ok((fixed!(0), fixed!(0)));
+        }
+        let (max_share_reserves_delta, success) =
+            self.calculate_max_share_reserves_delta(current_block_timestamp)?;
```
jalextowle, line 344: Should we call this `calculate_max_share_reserves_delta_safe`?

crates/hyperdrive-math/src/lp/math.rs
```diff
@@ -281,6 +301,652 @@ impl State {
...
+        //
+        // x_0 = w * (PV(0) / l)
+        let starting_present_value = self.calculate_present_value(current_block_timestamp)?;
+        let mut share_proceeds =
+            withdrawal_shares_total_supply.mul_div_down(starting_present_value, lp_total_supply);
+
+        // If the pool is net neutral, the initial guess is equal to the final
+        // result.
+        let net_curve_trade =
+            self.calculate_net_curve_trade_from_timestamp(current_block_timestamp)?;
```
jalextowle, line 501: Do we need to use the `safe` variant?

crates/hyperdrive-math/src/lp/math.rs
```diff
@@ -281,6 +301,652 @@ impl State {
...
+
+                // If the net curve trade is greater than or equal to the max
+                // bond amount, we can solve directly for the share proceeds.
+                let net_curve_trade = FixedPoint::from(U256::try_from(net_curve_trade)?);
+                if net_curve_trade >= max_bond_amount {
+                    // Solve the objective function directly assuming that it is
+                    // linear with respect to the share proceeds.
+
+                    let (share_proceeds, success) = updated_state
+                        .calculate_distribute_excess_idle_share_proceeds_net_long_edge_case(
```
jalextowle, line 582: Should we call this one `_safe`?

crates/hyperdrive-math/src/lp/math.rs
```diff
@@ -629,6 +1295,402 @@ mod tests {
...
+            present_state.info.long_average_maturity_time += current_block_timestamp;
+            present_state.info.short_average_maturity_time += current_block_timestamp;
+
+            // active_lp_total_supply and _withdrawal_shares_total_supply are just the token supplies.
+            // Neither are supplied in the PoolInfo so we need to make them here.  lp_total_supply is defined as:
+            // lp_total_supply = active_lp_total_supply + withdrawal_shares_total_supply - withdrawal_shares_ready_to_withdraw
+            // active_lp_total_supply = lp_total_supply - withdrawal_shares_total_supply + withdrawal_shares_ready_to_withdraw
+            // We clip withdrawal_shares_total_supply to ensure active_lp_total_supply doesn't underflow.
+            let active_lp_total_supply = present_state.lp_total_supply()
+                + present_state.withdrawal_shares_ready_to_withdraw();
```
jalextowle, line 1322: This should be `lp_total_supply + (withdrawal_share_total_supply - withdrawal_shares_ready_to_withdraw)`.

crates/hyperdrive-math/src/lp/math.rs
```diff
@@ -629,6 +1295,402 @@ mod tests {
...
+                .calculate_distribute_excess_idle(params, SHARE_PROCEEDS_MAX_ITERATIONS.into())
+                .call()
+                .await
+            {
+                Ok(expected) => {
+                    let (sol_withdrawal_shares_redeemed, sol_share_proceeds) = expected;
+                    println!(
+                        "\nsol_withdrawal_shares_redeemed  {:#?}",
+                        sol_withdrawal_shares_redeemed
+                    );
```
jalextowle, line 1394: We should probably remove these prints before merging.

jalextowle (commented):
It's looking good. Just a few more comments.

crates/hyperdrive-math/src/lp.rs
sentilesdal: I wanna follow up with this later so this PR is easier to review

crates/hyperdrive-math/src/lp/math.rs
sentilesdal: yeah it _is_ using a success flag

crates/hyperdrive-math/src/lp/math.rs
sentilesdal: yeah, using flag

crates/hyperdrive-math/src/lp/math.rs
sentilesdal: oh yeah, this is happening two lines below.  i do it piecemeal to make sure i don't underflow when i subtract withdrawal_shares_ready_to_withdraw.

crates/hyperdrive-math/src/lp/math.rs
sentilesdal: So this actually was the safe version, it was just mislabelled.  I updated the name and also added in the unsafe version which `distribute_excess_idle_withdrawal_shares_redeemed` uses.

PR #58, opened 2024-05-02
dynamic fuzz

dpaiton:
This PR is to fix an ongoing problem where we get intermittent fuzz tests failures in CI. The problem is that the tests don't get enough coverage from the default number of fuzz runs. However, we don't want to bump up the constants because then CI will take forever to run tests that we assume are fine. I fix this by increasing the number of fuzz runs on _new_ tests in a PR.

This is to be coupled with an AWS service that runs _all_ fuzz tests continuously. This will increase our coverage in general.

- adds a new test to test improved test testing
- adds a script to find all the tests
- adds a GH workflow to run new tests w/ hella fuzz


sentilesdal (approved):
looks good, clever!

PR #75, opened 2024-05-03
Update hyperdrive-wrappers to hyperdrive v1.0.5

sentilesdal:
# Resolved Issues

# Description
Commits tell a pretty good story, I'd recommend using them.  This PR gets us to hyperdrive version 1.0.5 which had some breaking changes.  The big one was fixing the deployments that required linking LPMath in the bytecode.  The wrappers all now have their own `deploy_linked_contract()` method that can do this for us.  The rest are in this list and a pretty well separated by commit.

- update version to 1.0.5
- remove `deploy()` aliases from build.rs since 1.0.5 renames the conflicting methods.
- add circuit breaker delta where needed
- remove target4 calls
- add `deploy_linked_contract()` method to the wrappers
- fix deployments that require LPMath linked reference.
- fix `Checkpoint` calls

# Review Checklists

- [ ] **Testing**


crates/hyperdrive-wrappers/Cargo.toml
```diff
@@ -10,19 +10,26 @@ build = "build.rs"
 
 # External dependencies
 ethers = "2.0.8"
+ethers-solc = "2.0.8"
```
jalextowle, line 13: Why is this a runtime dependency now? It should still only be used in `build.rs`

jalextowle (approved):
LGTM

crates/hyperdrive-wrappers/Cargo.toml
jalextowle: Nvm, I see where it is used

PR #79, opened 2024-05-05
add add_liquidity variant bindings

jrhea:
# Description
add add_liquidity variant bindings
# Review Checklists

- [ ] **Testing**

### DIFF
--- a/bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
+++ b/bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
@@ -126,7 +126,12 @@ impl HyperdriveState {
         let (result_fp1, result_fp2, result_fp3) = self
             .state
             .calculate_pool_deltas_after_add_liquidity(contribution_fp, true)
-            .unwrap();
+            .map_err(|err| {
+                PyErr::new::<PyValueError, _>(format!(
+                    "calculate_pool_deltas_after_add_liquidity returned the error: {:?}",
+                    err
+                ))
+            })?;
         let result1 = U256::from(result_fp1).to_string();
         let result2 = I256::from(result_fp2).to_string();
         let result3 = U256::from(result_fp3).to_string();
--- a/bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
+++ b/bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
@@ -116,6 +116,23 @@ impl HyperdriveState {
         Ok(result)
     }
 
+    pub fn calculate_pool_deltas_after_add_liquidity(
+        &self,
+        contribution: &str,
+    ) -> PyResult<(String, String, String)> {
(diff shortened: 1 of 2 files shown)


bindings/hyperdrivepy/src/hyperdrive_state_methods.rs
```diff
@@ -116,6 +116,23 @@ impl HyperdriveState {
...
+        &self,
+        contribution: &str,
+    ) -> PyResult<(String, String, String)> {
+        let contribution_fp = FixedPoint::from(U256::from_dec_str(contribution).map_err(|_| {
+            PyErr::new::<PyValueError, _>("Failed to convert contribution string to U256")
+        })?);
+        let (result_fp1, result_fp2, result_fp3) = self
+            .state
+            .calculate_pool_deltas_after_add_liquidity(contribution_fp, true)
+            .unwrap();
```
sentilesdal, line 129: instead of unwrapping, I would call `.map_error()` and map it to a PyError.  I think there is prior art for this in the file.

sentilesdal: Yeah do something like this:
```
            .map_err(|err| {
                PyErr::new::<PyValueError, _>(format!(
                    "Calculate_targeted_long_with_budget returned the error: {:?}",
                    err
                ))
            })?;
 ```
 
Fix the comment obv, but this will catch the error and forward it instead of panic-ing if the function fails for a known reason.

**Review by sentilesdal — APPROVED** [review]
lgtm, one change to map the erorr.   thx for getting rid of those unused fn args.

File: bindings/hyperdrivepy/src/hyperdrive_state_methods.rs (hunk 1 above)
Comment (jrhea) [reply bindings/hyperdrivepy/src/hyperdrive_state_methods.rs:129]: roger

File: crates/hyperdrive-math/src/lp.rs [hunk 2]
```diff
@@ -84,11 +83,7 @@ impl State {
 
     pub fn calculate_pool_deltas_after_add_liquidity(
         &self,
-        current_block_timestamp: U256,
         contribution: FixedPoint,
-        min_lp_share_price: FixedPoint,
-        min_apr: FixedPoint,
-        max_apr: FixedPoint,
```
ryangoree, line 91: nice

crates/hyperdrive-math/src/lp.rs
```diff
@@ -757,7 +752,7 @@ mod tests {
             alice
                 .advance_time(
                     rate,
-                    FixedPoint::from(config.checkpoint_duration) * fixed!(0.5e18),
+                    FixedPoint::from(config.checkpoint_duration).mul_down(fixed!(0.5e18)),
```
ryangoree, line 755: why this switch from operators?

ryangoree (approved):

crates/hyperdrive-math/src/lp.rs
jrhea: i just did it to ensure it matched solidity

PR #89, opened 2024-05-09
Fix/test calculate pool deltas after open short

jrhea:
`test_calculate_pool_deltas()` after open short by accounting for fees in shares


sentilesdal (approved):
looks good to me

PR #92, opened 2024-05-09
Fix cyclic dependency between `test-utils` and `hyperdrive-math` crates

sentilesdal:
# Description
hyperdrive-math was depending on test-utils and vice versa.  The culprit is the `Agent` struct in `test_utils` that required a lot of things in `hyperdrive-math` for the `Agent` methods.  `hyperdrive-math` then imported the Agent to run tests.  

This PR removes the cyclic dependency by extending the `Agent` struct in `hyperdrive-math` with it's own `HyperdriveMathAgent` in test_utils/agent.rs.  The agent module is `#[cfg(test)]` decorated so it only compiles for tests and isn't included in the release binary.  

One caveat to this approach though is that other crates will not be able to import the HyperdriveMathAgent, including the tests/ directory, so I had to move tests/integration_tests.rs to src/test_utils/integration_tests.rs

# Review Checklists

- [ ] **Testing**


crates/hyperdrive-math/Cargo.toml
```diff
@@ -20,6 +15,9 @@ rand = "0.8.5"
 fixed-point = { path = "../fixed-point" }
 fixed-point-macros = { path = "../fixed-point-macros" }
 hyperdrive-wrappers = { path = "../hyperdrive-wrappers" }
+test-utils = { path = "../test-utils" }
+rand_chacha = "0.3.1"
+tracing = "0.1"
```
ryangoree, line 20: Can we move these to `dev-dependencies`

crates/hyperdrive-math/src/test_utils/agent.rs
```diff
@@ -0,0 +1,455 @@
...
+use rand_chacha::ChaCha8Rng;
+use test_utils::{
+    agent::{Agent, ContractCall, TxOptions},
+    chain::ChainClient,
+};
+use tracing::instrument;
+
+use crate::State;
+
+pub trait HyperdriveMathAgent {
```
ryangoree, line 17: Does the trait have to be defined first to implement the methods? Could we just implement the trait directly and avoid the duplication?

crates/test-utils/src/agent.rs
```diff
@@ -21,29 +21,17 @@ use hyperdrive_wrappers::wrappers::{
...
-    RemoveLiquidity(FixedPoint),
-    RedeemWithdrawalShares(FixedPoint),
-    OpenLong(FixedPoint),
-    CloseLong(FixedPoint, FixedPoint),
-    OpenShort(FixedPoint),
-    CloseShort(FixedPoint, FixedPoint),
-}
-
-#[derive(Clone, Default)]
+#[derive(Clone, Default, Getters)]
```
ryangoree, line 28: What is `Getters` doing here? I see the crate let's you generate getters on private methods, but I'm not seeing the `#[getset...]` attribute added anywhere.

ryangoree (approved):

crates/test-utils/src/agent.rs
sentilesdal: Thanks, I added this and then removed it because it wasn't working as expected so I made the Wallet members public instead.

crates/hyperdrive-math/Cargo.toml
sentilesdal: yeah we can now that Agent is only compiled for tests!

crates/hyperdrive-math/src/test_utils/agent.rs
sentilesdal: Yup you need it. If you try to do this in hyperdrive-math when Agent is defined in test_utils:
```rust
impl Agent<ChainClient<LocalWallet>, ChaCha8Rng> {
```
You'll get this:
```
cannot define inherent `impl` for a type outside of the crate where the type is defined
define and implement a trait or new type insteadrustc[Click](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B1%5D?1#file%3A%2F%2F%2FUsers%2Fmatthewbrown%2Fsrc%2Fdelv%2Fhyperdrive-rs%2Fcrates%2Fhyperdrive-math%2Fsrc%2Ftest_utils%2Fagent.rs)```

File: crates/test-utils/src/agent.rs [hunk 4]
```diff
@@ -187,255 +174,16 @@ impl Agent<ChainClient<LocalWallet>, ChaCha8Rng> {
...
-                    })
-                    .collect::<Vec<_>>();
-            logs[0].clone()
-        };
-        self.wallet.base += log.base_amount.into();
-
-        Ok(log.base_amount.into())
-    }
-
     /// LPs ///
```
dpaiton, line 187: dead comment?

crates/test-utils/src/chain/deploy.rs
```diff
@@ -55,6 +54,38 @@ where
     Ok(u256)
 }
 
+pub fn calculate_time_stretch(rate: FixedPoint, position_duration: FixedPoint) -> FixedPoint {
```
dpaiton, line 57: We use this in `agent0` and so it should not be moved without updating PRs elsewhere.
https://github.com/delvtech/agent0/blob/main/src/agent0/ethpy/hyperdrive/interface/_mock_contract.py#L25

dpaiton: Wait I'm sorry, it looks like you did not delete the version in `hyperdrive-math/src/utils.rs`. Why did you duplicate the code here instead of using the imported version? Is this part of the circular dependency thing?

dpaiton (approved):
minor confusion/questions but in general this is a great improvement! Thanks

crates/test-utils/src/agent.rs
```diff
@@ -187,255 +174,16 @@ impl Agent<ChainClient<LocalWallet>, ChaCha8Rng> {
         })
     }
 
+    pub fn client(&self) -> Arc<ChainClient<LocalWallet>> {
+        self.client.clone()
+    }
+
     pub fn hyperdrive(&self) -> &IHyperdrive<ChainClient<LocalWallet>> {
```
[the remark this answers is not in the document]
sentilesdal: No I didn't go that far, its more of a barebones hyperdrive agent.  Anything that needs `hyperdrive-math` is now in that crate.  I also really like the idea of having a protocol agnostic barebones test-utils with agents but I don't think now is the time.

crates/test-utils/src/agent.rs
```diff
@@ -187,255 +174,16 @@ impl Agent<ChainClient<LocalWallet>, ChaCha8Rng> {
...
         &self.hyperdrive
     }
 
     /// Longs ///
```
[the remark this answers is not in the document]
sentilesdal: yes ty

crates/test-utils/src/agent.rs
```diff
@@ -457,7 +205,7 @@ impl Agent<ChainClient<LocalWallet>, ChaCha8Rng> {
 
         // Initialize the pool and record the LP shares that were received in the wallet.
         let log = {
-            let tx = ContractCall_(self.hyperdrive.initialize(
```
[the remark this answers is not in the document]
sentilesdal: Keeping here because it doesn't require any hyperdrive-math stuff.

crates/test-utils/src/chain/deploy.rs
sentilesdal: yeah this would make another circular dependency :P.  This isn't great I know, but one problem at a time I think.

sentilesdal: I'll add a comment to the top of this one.

crates/test-utils/src/agent.rs
sentilesdal: ya

PR #95, opened 2024-05-13
fee overhaul & targeted long fix

### LINKED ISSUE #43: Avoid redundant calculations for `calculate_close_short_flat_plus_curve`
https://github.com/delvtech/hyperdrive/pull/916 adds a couple of negative interest checks, which requires a few internal variables used by `calculate_close_short_flat_plus_curve`. Additionally, a new `close_short_governance_fee` also includes a potentially redundant `close_short_curve_fee`. Further work is needed to remove redundant calculations of these internal variables.

dpaiton:
# Description
This PR is the result of an audit of the fee functions. The following changes were made (one commit per bullet):

1. revert the rounding adjustment from #97. We realized that the bug actually lies in the solidity _test utils_. This has been fixed so I reverted the rounding behavior to match.
2. update docstrings to use a more consistent format. I introduce a convention of my own to use upper-case $\Phi_{c,ol}(...)$ for the fee function and lower-case $\phi_{c}$ for the fee multiplier constant. The subscript on the fee function is the fee type $\in c, g, f$ for curve, gov, and flat; as well as the trade type $\in ol, cl, os, cs$ for opening or closing a long or short. This replaces the inconsistent method of using e.g. `c(x)` at some times and `curve_fee` at other times, or using repeat symbols for different equations.
3. add `maybe_curve_fee` arg to gov fee calculations to fix #43 and reduce computation.
4. change the rounding behavior to match solidity (note it's matching the eqs in `internal/HyperdriveBase.sol`
5. the above changes revealed a unit issue in `calculate_targeted_long` where the gov fee (and derivative) was not converted to shares. I updated my [overleaf doc](https://www.overleaf.com/read/jkngspgyqzkf#1de6ce), the docstrings, and the code itself.

# Review Checklists

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/long/fees.rs
+++ b/crates/hyperdrive-math/src/long/fees.rs
@@ -27,10 +27,16 @@ impl State {
     /// $$
     /// \Phi_{g,ol}(\Delta x) = \phi_g \cdot p \cdot \Phi_{c,ol}(\Delta x)
     /// $$
-    pub fn open_long_governance_fee(&self, base_amount: FixedPoint) -> FixedPoint {
-        self.governance_lp_fee()
-            * self.calculate_spot_price()
-            * self.open_long_curve_fee(base_amount)
+    pub fn open_long_governance_fee(
+        &self,
+        base_amount: FixedPoint,
+        maybe_curve_fee: Option<FixedPoint>,
+    ) -> FixedPoint {
+        let curve_fee = match maybe_curve_fee {
+            Some(maybe_curve_fee) => maybe_curve_fee,
+            None => self.open_long_curve_fee(base_amount),
+        };
+        self.governance_lp_fee() * self.calculate_spot_price() * curve_fee
     }
 
     /// Calculates the curve fee paid when closing longs for a given bond amount.
--- a/crates/hyperdrive-math/src/long/fees.rs
+++ b/crates/hyperdrive-math/src/long/fees.rs
@@ -5,36 +5,44 @@ use fixed_point_macros::fixed;
 use crate::State;
 
 impl State {
-    /// Calculates the curve fee paid by longs for a given base amount.
+    /// Calculates the curve fee paid when opening longs with a given base amount.
     ///
-    /// The curve fee $c(x)$ paid by longs is paid in bonds and is given by:
+    /// The open long curve fee, $\Phi_{c,ol}(\Delta x)$, is paid in bonds and
+    /// is given by:
     ///
     /// $$
-    /// c(x) = \phi_c \cdot \left( \tfrac{1}{p} - 1 \right) \cdot x
+    /// \Phi_{c,ol}(\Delta x) = \phi_c \cdot \left( \tfrac{1}{p} - 1 \right) \cdot \Delta x
     /// $$
-    pub fn open_long_curve_fees(&self, base_amount: FixedPoint) -> FixedPoint {
+    pub fn open_long_curve_fee(&self, base_amount: FixedPoint) -> FixedPoint {
         self.curve_fee()
             * ((fixed!(1e18) / self.calculate_spot_price()) - fixed!(1e18))
             * base_amount
     }
 
-    /// Calculates the governance fee paid by longs for a given base amount.
+    /// Calculates the governance fee paid when opening longs with a given base amount.
     ///
-    /// Unlike the [curve fee](long_curve_fee) which is paid in bonds, the
-    /// governance fee is paid in base. The governance fee $g(x)$ paid by longs
+    /// The open long governance fee, $\Phi_{g,ol}(\Delta x)$, is paid in base and
     /// is given by:
     ///
     /// $$
-    /// g(x) = \phi_g \cdot p \cdot c(x)
+    /// \Phi_{g,ol}(\Delta x) = \phi_g \cdot p \cdot \Phi_{c,ol}(\Delta x)
     /// $$
     pub fn open_long_governance_fee(&self, base_amount: FixedPoint) -> FixedPoint {
         self.governance_lp_fee()
             * self.calculate_spot_price()
-            * self.open_long_curve_fees(base_amount)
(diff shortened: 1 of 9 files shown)


crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -229,68 +229,79 @@ impl State {
     /// is equal to
     ///
```
dpaiton, line 230: strictly formatting change here

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -229,68 +229,79 @@ impl State {
...
+    ///     {y - \Delta y} \right)^{t_{s}}
     /// $$
     ///
     /// where $t_{s}$ is the time stretch constant and $z_{e,0}$ is the initial
     /// effective share reserves, and $\zeta$ is the zeta adjustment.
     /// The zeta adjustment is constant when opening a long, i.e.
     /// $\Delta \zeta = 0$, so we drop the subscript. Equivalently, for some
-    /// amount of `delta_base`$= x$ provided to open a long, we can write:
+    /// amount of `delta_base`$= \Delta x$ provided to open a long, we can write:
     ///
```
dpaiton, line 242: math change below fixes typo; should divide gov fee by vault share price

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -229,68 +229,79 @@ impl State {
...
     /// $$
     ///
-    /// where $g(x)$ is the [open_long_governance_fee](long::fees::open_long_governance_fee),
-    /// $y(x)$ is the [long_amount](long::open::calculate_open_long),
-    ///
+    /// where $\Phi_{g,ol}(\Delta x)$ is the [open_long_governance_fee](long::fees::open_long_governance_fee),
+    /// $y(\Delta x)$ is the [bond_amount](long::open::calculate_open_long),
     ///
     /// To compute the derivative, we first define some auxiliary variables:
     ///
```
dpaiton, line 255: same fix below; should be dividing gov fee by vault share price

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -229,68 +229,79 @@ impl State {
...
+    ///     &= \mu \left( z_{e,0} + \frac{\Delta x}{c} - \frac{\Phi_{g,ol}(\Delta x)}{c} \right) \\
+    /// b(\Delta x) &= y_0 - y(\Delta x) \\
+    /// v(\Delta x) &= \frac{a(\Delta x)}{b(\Delta x)}
     /// $$
     ///
-    /// and thus $p(x) = v(x)^t_{s}$. Given these, we can write out intermediate derivatives:
+    /// and thus $p(\Delta x) = v(\Delta x)^{t_{s}}$.
+    /// Given these, we can write out intermediate derivatives:
     ///
     /// $$
```
dpaiton, line 266: $a'(x)$ had an error after propagating the above fixes. Other changes are cosmetic.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -348,7 +360,6 @@ mod tests {
     use super::*;
     use crate::test_utils::agent::HyperdriveMathAgent;
 
-    #[ignore]
```
dpaiton, line 351: I ran the test locally 3 times and it passed every time.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -229,67 +229,79 @@ impl State {
...
         bond_amount: FixedPoint,
     ) -> Result<FixedPoint> {
-        // g'(x)
+        // g'(x) = \phi_g \phi_c (1 - p_0)
         let gov_fee_derivative = self.governance_lp_fee()
             * self.curve_fee()
             * (fixed!(1e18) - self.calculate_spot_price());
 
-        // a(x) = mu * (z_{e,0} + x/c - g(x))
+        // a(x) = mu * (z_{e,0} + 1/c (x - g(x))
```
jrhea, line 288: nice catch

crates/hyperdrive-math/src/long/targeted.rs
jrhea: image

jrhea (approved):
lgtm

crates/hyperdrive-math/src/long/fees.rs
```diff
@@ -5,36 +5,54 @@ use fixed_point_macros::fixed;
...
     /// $$
-    pub fn open_long_curve_fees(&self, base_amount: FixedPoint) -> FixedPoint {
-        self.curve_fee()
-            * ((fixed!(1e18) / self.calculate_spot_price()) - fixed!(1e18))
-            * base_amount
+    pub fn open_long_curve_fee(&self, base_amount: FixedPoint) -> FixedPoint {
+        // NOTE: Round up to overestimate the curve fee.
+        (fixed!(1e18).div_up(self.calculate_spot_price()) - fixed!(1e18))
+            .mul_up(self.curve_fee())
+            .mul_up(base_amount)
```
ryangoree, line 20: How does this work? And the units returned is in bonds?

Looking at this: https://github.com/delvtech/hyperdrive/blob/33982bba72d861b3b90cb0b4083fec30accb649f/contracts/src/internal/HyperdriveBase.sol#L773-L776

I would think we'd have to do something like this to get the curve fee in bonds:

```
curve_fee
  .mul_up(1 - spot_price)
  .mul_up(base_amount.div_up(spot_price))
  .mul_up(normalize_time_remaining)
```

jrhea: For open long we use this: https://github.com/delvtech/hyperdrive/blob/33982bba72d861b3b90cb0b4083fec30accb649f/contracts/src/internal/HyperdriveBase.sol#L718

Issue #88, opened 2024-05-09
Remaining Flaky Tests


jrhea:
We are ignoring failures with these tests currently. This needs to be addressed ASAP

- [x]     `long::targeted::tests::test_calculate_targeted_long_with_budget`
- [x]     `short::max::tests::fuzz_calculate_absolute_max_short_execute`
- [x]     `short::max::tests::fuzz_calculate_max_short_no_budget`

dpaiton:
https://github.com/delvtech/hyperdrive-rs/pull/95 resolves the remaining flaky tests, but unfortunately introduces a new one:

- [ ] `long::max::fuzz_calculate_max_long`

I'm investigating that now

dpaiton:
another one:

```
failures:

---- short::max::tests::fuzz_calculate_max_short stdout ----
thread 'short::max::tests::fuzz_calculate_max_short' panicked at crates/hyperdrive-math/src/short/max.rs:723:17:
expected (base=34326359.285477447501009068) < (budget=35401424.462665845134388594) * 0.001000000000000000 = 35401.424462665845134388


failures:
    short::max::tests::fuzz_calculate_max_short
```

dpaiton:
> #95 resolves the remaining OP flaky tests, but unfortunately introduces a new one:
> 
> * [x]  `long::max::fuzz_calculate_max_long`
> 
> I'm investigating that now

resolved by https://github.com/delvtech/hyperdrive-rs/pull/101

dpaiton:
> another one:
> 
> ```
> failures:
> 
> ---- short::max::tests::fuzz_calculate_max_short stdout ----
> thread 'short::max::tests::fuzz_calculate_max_short' panicked at crates/hyperdrive-math/src/short/max.rs:723:17:
> expected (base=34326359.285477447501009068) < (budget=35401424.462665845134388594) * 0.001000000000000000 = 35401.424462665845134388
> 
> 
> failures:
>     short::max::tests::fuzz_calculate_max_short
> ```

I ran the fuzz test 4 * 1000 times without failure against the branch for this PR https://github.com/delvtech/hyperdrive/pull/1028

I'm going to close the issue; we can re-open one later if we find new flaky tests after v1.0.7 hits with #1028 merged

PR #97, opened 2024-05-14
Fix `open_short_curve_fee`

ryangoree:
# Resolved Issues

Related to #88 

# Description

We were using `mul_up` where the contracts use `mulDown`: https://github.com/delvtech/hyperdrive/blob/33982bba72d861b3b90cb0b4083fec30accb649f/test/utils/HyperdriveUtils.sol#L1362

# Review Checklists

- [ ] **Testing**


jalextowle (approved):
As mentioned offline, we may want to wait on the Solidity fixes. I’ll approve in case this makes the fixes tonight/tomorrow easier, but we’ll want to revert this after v1.0.7 lands.

Issue #22, opened 2024-03-20
implement `calculate_remove_liquidity`


ryangoree:
Add a function to the Rust SDK that calculates the amount of `proceeds` and `withdrawal_shares` received for removing liquidity. Traders and UIs could use this to preview their returns without hitting the chain.

sentilesdal:
Delayed a bit by the 1.0.5 upgrade but,

Two landed:
#7, #11
One ready to go:
#17 - ShouldShortCircuit and LongEdgeCase helpers

Not in PR, but code written:
DistributeExcessIdleShortProceeds, DistributeExcessIdle

sentilesdal:
Delayed a bit by the 1.0.5 upgrade but,

Two landed:
https://github.com/delvtech/hyperdrive-rs/pull/7, https://github.com/delvtech/hyperdrive-rs/pull/11
One ready to go:
https://github.com/delvtech/hyperdrive-rs/pull/17 - `distribute_excess_idle`
One draft PR:
https://github.com/delvtech/hyperdrive-rs/pull/100 - `calculate_remove_liquidity`

sentilesdal:
Three landed:
https://github.com/delvtech/hyperdrive-rs/pull/7, https://github.com/delvtech/hyperdrive-rs/pull/11
https://github.com/delvtech/hyperdrive-rs/pull/17 - distribute_excess_idle
One Ready to Go:
https://github.com/delvtech/hyperdrive-rs/pull/100 - calculate_remove_liquidity

sentilesdal:
#100 needs https://github.com/delvtech/hyperdrive/pull/1049 to land first in order to have the exact results

PR #100, opened 2024-05-15
Add `calculate_remove_liquidity`

sentilesdal:
# Resolved Issues
https://github.com/delvtech/hyperdrive-rs/issues/22

# Description
Adds `calculate_remove_liquidity` and a test.

# Review Checklists

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/lp/remove.rs
+++ b/crates/hyperdrive-math/src/lp/remove.rs
@@ -1,10 +1,9 @@
-use crate::State;
-
 use ethers::types::{I256, U256};
 use eyre::{eyre, Result};
 use fixed_point::{fixed, FixedPoint};
 
 use super::math::SHARE_PROCEEDS_MAX_ITERATIONS;
+use crate::State;
 
 impl State {
     /// Allows an LP to burn shares and withdraw from the pool.
--- a/crates/hyperdrive-math/src/lp/remove.rs
+++ b/crates/hyperdrive-math/src/lp/remove.rs
@@ -18,7 +18,7 @@ impl State {
         min_output_per_share: FixedPoint,
         minimum_transaction_amount: FixedPoint,
         as_base: bool,
-    ) -> Result<(FixedPoint, FixedPoint)> {
+    ) -> Result<(FixedPoint, FixedPoint, State)> {
         // Ensure that the amount of LP shares to remove is greater than or
         // equal to the minimum transaction amount.
         if lp_shares < minimum_transaction_amount {
@@ -28,8 +28,8 @@ impl State {
         // Get vault share price and proceed without checkpoint.
         let vault_share_price = self.vault_share_price();
 
-        let mut state = self.clone();
         // Burn the LP's shares.
+        let mut state = self.clone();
         state.info.lp_total_supply -= lp_shares.into();
         let active_lp_total_supply = active_lp_total_supply - lp_shares;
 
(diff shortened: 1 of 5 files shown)


dpaiton:
would you mind adding the python bindings?

crates/hyperdrive-math/src/lp/remove.rs
```diff
@@ -1,6 +1,312 @@
...
+        // Get vault share price and proceed without checkpoint.
+        let vault_share_price = self.vault_share_price();
+
+        let mut state = self.clone();
+        // Burn the LP's shares.
+        state.info.lp_total_supply -= lp_shares.into();
+        let active_lp_total_supply = active_lp_total_supply - lp_shares;
+
+        // Mint an equivalent amount of withdrawal shares.
+        let withdrawal_shares_total_supply = withdrawal_shares_total_supply + lp_shares;
```
dpaiton, line 34: you can declare the variable as `mut` in the fn signature and then use `+=`. I am not sure which is better, tbh, but the latter pattern is what we use e.g. in [yieldspace.rs](https://github.com/delvtech/hyperdrive-rs/blob/dpaiton/calc-open-short/crates/hyperdrive-math/src/yield_space.rs#L274) and [short/open.rs](https://github.com/delvtech/hyperdrive-rs/blob/dpaiton/calc-open-short/crates/hyperdrive-math/src/short/open.rs#L43).

crates/hyperdrive-math/src/lp/remove.rs
```diff
@@ -1,6 +1,312 @@
...
+            current_block_timestamp,
+            active_lp_total_supply,
+            withdrawal_shares_total_supply,
+            lp_shares,
+            vault_share_price,
+            min_output_per_share,
+        )?;
+        let withdrawal_shares = lp_shares - withdrawal_shares_redeemed;
+
+        Ok((proceeds, withdrawal_shares))
```
dpaiton, line 47: Elsewhere (e.g. [short/open](https://github.com/delvtech/hyperdrive-rs/blob/dpaiton/calc-open-short/crates/hyperdrive-math/src/short/open.rs#L160-L177) and [long/open](https://github.com/delvtech/hyperdrive-rs/blob/dpaiton/calc-open-short/crates/hyperdrive-math/src/long/open.rs#L53-L76)) we have been separating functions that compute the new state and new deltas. This case is a little tricky... You permute state as part of the process (adjusting lp total supply), which I assume is then considered when calculating redeem withdrawal shares.

Do you think it makes sense to include the new `state` in the return tuple, or you could return a `lp_supply_delta` variable to be used in a later `calculate_pool_state_after_remove_liquidity` function. Or don't add an extra variable and re-compute it in a dedicated `calculate_pool_deltas*` fn.

Thoughts?

crates/hyperdrive-math/src/lp/remove.rs
```diff
@@ -1,6 +1,312 @@
...
+        // to avoid unnecessary reverts. We exit early if the user has no shares
+        // available to redeem.
+        let mut withdrawal_shares_redeemed = withdrawal_shares;
+        if withdrawal_shares_redeemed > ready_to_withdraw {
+            withdrawal_shares_redeemed = ready_to_withdraw;
+        }
+        if withdrawal_shares_redeemed == fixed!(0) {
+            return Ok((fixed!(0), fixed!(0)));
+        }
+
```
dpaiton, line 91: solidity burns shares from the user here; was that skipped on purpose?

crates/hyperdrive-math/src/lp/remove.rs
```diff
@@ -1,6 +1,312 @@
...
+        // If there is no excess idle, then there is nothing to distribute.
+        let idle = self.calculate_idle_share_reserves();
+        if idle == fixed!(0) {
+            return Ok((fixed!(0), fixed!(0), true));
+        }
+
+        // Calculate the amount of withdrawal shares that should be redeemed
+        // and their share proceeds.
+        let (withdrawal_shares_redeemed, share_proceeds) = self.calculate_distribute_excess_idle(
```
dpaiton, line 137: solidity has `_distributeExcessIdleSafe()`, which I believe does the actual distribution & includes checks, and `calculateDistributeExcessIdle()` to compute the `withdrawalSharesRedeemed` and `shareProceeds`.

Here in rust you're renaming the first function to `calculate_distribute_excess_idle_safe` and the second to `calculate_distribute_excess_idle`. I think these names are confusing -- it implies that they're the same function and one is the "safe" version of it. 

This is a case where we should depart from exactly copying solidity in favor of rust readability. I think you should rename one of these functions.

dpaiton (approved):
lgtm but @jalextowle  said he was going to review as well so you should wait for confirmation from him before merging

crates/hyperdrive-math/src/lp/remove.rs
sentilesdal: It's been my preference to not use `mut` when possible and just redeclare.  Avoiding mutability is kinda Rust's bag.

crates/hyperdrive-math/src/lp/remove.rs
sentilesdal: skipped on purpose, rust doesn't have that state.  these are `calculate_x` functions so as long as we return a correct calculation we can ignore resulting state updates.  Also, by state I mean contract state, i.e. things like user token balances, not the `State` object we pass around.

crates/hyperdrive-math/src/lp/remove.rs
```diff
@@ -1,6 +1,312 @@
...
+            match result {
+                Ok((shares, withdrawal_shares)) => {
+                    // Assert base redeemed results match between rust and solidity
+                    // within a tolerance.
+                    let sol_base_result = I256::try_from(ending_base - starting_base)?;
+                    let vault_share_price = state.vault_share_price();
+                    let rust_base_reesult = I256::try_from(shares * vault_share_price)?;
+                    let base_diff =
+                        FixedPoint::try_from((sol_base_result - rust_base_reesult).abs())? / shares;
+                    assert!(base_diff < fixed!(0.0000001e18));
```
jalextowle, line 290: I'm not sure why we need this. Shouldn't this be perfectly accurate?

crates/hyperdrive-math/src/lp/remove.rs
```diff
@@ -1,6 +1,312 @@
...
+
+                    // Assert withdrawal shares results match between rust and
+                    // solidity within a tolerance.
+                    let sol_ws_result =
+                        I256::try_from(ending_withdrawal_shares - starting_withdrawal_shares)?;
+                    let rust_ws_result = I256::try_from(withdrawal_shares)?;
+                    let withdrawal_shares_diff =
+                        FixedPoint::try_from((sol_ws_result - rust_ws_result).abs())? / shares;
+                    assert!(withdrawal_shares_diff < fixed!(0.0000001e18));
```
jalextowle, line 299: Same here

crates/hyperdrive-math/src/lp/remove.rs
dpaiton: oh ok, I didn't know that. Should we change those examples I gave you then?

crates/hyperdrive-math/src/lp/remove.rs
dpaiton: ok cool thanks for clarifying :)

crates/hyperdrive-math/src/lp/remove.rs
sentilesdal: this is removed, exact now!

crates/hyperdrive-math/src/lp/remove.rs
sentilesdal: exact now!

crates/hyperdrive-math/src/lp/remove.rs
sentilesdal: I wouldn't go out of your way but in general we should avoid mutability unless needed for a for-loop scenario etc.

crates/hyperdrive-math/src/lp/remove.rs
```diff
@@ -1,6 +1,312 @@
...
+        if proceeds < min_output_per_share.mul_up(withdrawal_shares_redeemed) {
+            return Err(eyre!("Output limit not met"));
+        }
+
+        Ok((proceeds, withdrawal_shares_redeemed))
+    }
+
+    /// Distribute as much of the excess idle as possible to the withdrawal
+    /// pool while holding the LP share price constant.
+    fn calculate_distribute_excess_idle_safe(
```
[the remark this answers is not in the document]
sentilesdal: It seems like the pattern is to call things 'calculate' because we aren't updating any state, just returning values.

crates/hyperdrive-math/src/lp/remove.rs
sentilesdal: If there's a need for it we can, until then I'd say leave as is?

crates/hyperdrive-math/src/lp/remove.rs
sentilesdal: > This is a case where we should depart from exactly copying solidity in favor of rust readability. I think you should rename one of these functions.

This was my feeling as well, but honestly it's too much of a pain to maintain two different implementations for too little gain.  I tore my hair out trying to make things more readable only to revert to a more mirrored approach.  

I can rename this to whatever we want though, it's 'safe' because of the extra checks, how about renaming to get rid of the 'calculate' part and return updated `State` along with the updated values?

crates/hyperdrive-math/src/lp/remove.rs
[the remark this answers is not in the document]
sentilesdal: Ok actually yeah I'll remove the 'calculate' part and return updated state as a part of this function.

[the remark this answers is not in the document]
jalextowle: Ok sounds good. I'll do an early approve, but make sure you address this before merging.

jalextowle (approved):
LGTM

crates/hyperdrive-math/src/lp/remove.rs
sentilesdal: Ok I actually added this in because it helped with some extra test scenarios.

Issue #45, opened 2024-03-19
Fix `calculate_max_long` and `calculate_max_short` for accuracy


jalextowle:
A lot has changed in Hyperdrive over the last few months. We should make sure that the math used in these functions is actually correct and fix any discrepancies that we find. This should address any of the lingering known issues in the SDK.

wakamex:
crash on calc max short with a high fixed rate of 100%

failing test here: https://github.com/delvtech/hyperdrive-rs/pull/19

I noticed our fuzz ranges seem too narrow. should we do all of them up to 100% fixed rate? also in some cases fixed rate is not fuzzed.

wakamex:
crash trying to open long over 70% of the calculated max long after adding liquidity https://github.com/delvtech/agent0/commit/7434b9621eef5f10ba1b4ed29f23611c3cd4b6e7

dpaiton:
Related to https://github.com/delvtech/hyperdrive-rs/issues/42

wakamex:
copied from @slundqui's issue:

`absolute_max_long` is failing due to `target_share_reserves < effective_share_reserves` with `arithmetic operation overflow` here:
https://github.com/delvtech/hyperdrive/blob/1576e4c4368dd9c21b2df60274f223271e793d46/crates/hyperdrive-math/src/long/max.rs#L214

This is the pool state that caused this issue:
```
pool_config = {
   base_token='0x610178dA211FEF7D417bC0e6FeD39F05609AD788',
   vault_shares_token='0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e',
   linker_factory='0x5FbDB2315678afecb367f032d93F642f64180aa3',
   linker_code_hash=b'\x91\xa6<\xce\xc3\xbbXQ\xa0\x88\xc4G\xd5\xd3@`4AP\xbe+q\xac`\xc1\x97\x82\x7f\xd1HIL',
   initial_vault_share_price=FixedPoint("1.0"), minimum_share_reserves=FixedPoint("0.206282109669978528"),
   minimum_transaction_amount=FixedPoint("2.4934863032025728"),
   position_duration=30801600,
   checkpoint_duration=82800,
   time_stretch=FixedPoint("0.168128366020933774"),
   governance='0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
   fee_collector='0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266',
[...]

Issue #20, opened 2024-03-19
Improved Testing of the Rust SDK


jalextowle:
There were some issues reported a while ago by @slundqui. We'll need to review these calculations to see if we can find the source of the issue. We should also take the time to improve the test coverage. The current testing methodology is pretty good, but we don't test with different configurations, which really limits the robustness of the testing.

We should differentially test every rust function against Solidity.

dpaiton:
@ryangoree 
> Alternatively the FixedPoint library could be refactored to handle these cases internally. 

I think this is the move. We should refactor fixedpoint to return results instead of panic https://github.com/delvtech/hyperdrive-rs/issues/99

This would clean up **a lot** of testing code.

dpaiton:
https://github.com/delvtech/hyperdrive-rs/issues/127

dpaiton:
Updating test chain to support fuzzing over multiple states: https://github.com/delvtech/hyperdrive-rs/issues/128

dpaiton:
panic / result issue resolved as much as we want by https://github.com/delvtech/hyperdrive-rs/pull/113

dpaiton:
closing in favor of new smaller issues

PR #101, opened 2024-05-15
Fix fuzz max long test

dpaiton:
# Resolved Issues
Partially resolves these two:
https://github.com/delvtech/hyperdrive-rs/issues/88
https://github.com/delvtech/hyperdrive-rs/issues/45

# Description
The fee rounding behavior was adjusted recently to more closely match solidity, but the order of operations was not exactly the same. This caused an occasional slight variation in output (usually off by 1e1), that compounded in situations where we were iteratively assessing fees such as `max_long`. In that case, the error would grow to as much as 1e5.

During my effort to find this bug I
- modified some math to use the e.g. `.mul_down` syntax instead of `*` where it was sufficiently complicated to require this to easily pattern match against Solidity.
- modified `calculate_max_long` to return errors instead of panic. This makes checks in the tests ugly, but that is a temporary problem until we purge the panics all-together (https://github.com/delvtech/hyperdrive-rs/issues/20)
- purged the use of `traced` and `trace_test` which is helpful for debugging but was not actively being used and can slow down tests.

# Review Checklists

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/long/max.rs
+++ b/crates/hyperdrive-math/src/long/max.rs
@@ -1,4 +1,5 @@
 use ethers::types::I256;
+use eyre::{eyre, Result};
 use fixed_point::FixedPoint;
 use fixed_point_macros::{fixed, int256};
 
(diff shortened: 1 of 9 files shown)


crates/hyperdrive-math/src/long/max.rs
```diff
@@ -563,15 +562,25 @@ mod tests {
...
-            let state = rng.gen::<State>();
+            // Snapshot the chain.
+            let id = chain.snapshot().await?;
+
+            // Gen a random state.
+            let mut state = rng.gen::<State>();
+
+            // Make sure maturity times are in the future.
+            let current_block_timestamp = alice.now().await?;
+            state.info.long_average_maturity_time += current_block_timestamp;
```
sentilesdal, line 577: I don't think this matters here since calculate_max_long doesn't know about the current block timestamp.

sentilesdal (approved):
nice!

PR #105, opened 2024-05-17
Refactor crates to remove cyclic dependencies

ryangoree:
# Resolved Issues

Related to #92

# Description

This refactors the crates to remove all cyclic dependencies. Below is an illustration of the dependency tree before and after with a few notes:

hyperdrive-rs-dep-trees

# Review Checklists

- [ ] **Testing**


crates/hyperdrive-wrappers/build.rs
```diff
@@ -99,13 +101,14 @@ fn main() -> Result<()> {
     }
 
     // Compile the contracts.
+    // Command::new("forge")
```
sentilesdal: dead code?

crates/fixed-point/src/lib.rs
```diff
@@ -555,7 +553,7 @@ mod tests {
 
         // Fuzz the rust and solidity implementations against each other.
         let mut rng = thread_rng();
-        for _ in 0..*FAST_FUZZ_RUNS {
+        for _ in 0..1000 {
```
sentilesdal, line 570: leaving as 1000 on purpose?

crates/hyperdrive-test-utils/src/chain/deploy.rs
```diff
@@ -1042,6 +1060,8 @@ mod tests {
         chain.deal(ALICE.address(), uint256!(100_000e18)).await?;
         let client = chain.client(ALICE.clone()).await?;
 
+        println!("HERE");
```
sentilesdal, line 1063: println

crates/hyperdrive-test-utils/src/constants.rs
```diff
@@ -0,0 +1,17 @@
+use std::env;
+
+pub use test_utils::constants::*;
+
+lazy_static! {
```
sentilesdal, line 5: very nice, i need to do this in some of the lp files.

crates/hyperdrive-test-utils/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#![allow(clippy::comparison_chain)]
```
sentilesdal, line 1: ooo what's this?

sentilesdal (approved):
this is amazing, nice work.

crates/fixed-point/src/lib.rs
ryangoree: Nope! Good catch

crates/hyperdrive-test-utils/src/lib.rs
ryangoree: This is left over from before. It prevents clippy from complaining about using `if` statements in comparison chains that could be rewritten with `match` or `cmp`, but I just rant lint without it locally and it passed, so I don't think it's needed anymore 👍

sentilesdal: oh yeah, Ive seen those warnings, they're annoying.

PR #112, opened 2024-05-22
Fix calculate_implied_rate to return an annualized rate and scale tolerance up with implied_rate

MazyGio:
# Description
`calculate_implied_rate` was meant to return an annualized rate, but it was returning an HPR instead. This fixes it to return an annualized rate, and adds helper functions for turning annualized rates into HPR.

This also makes the test tolerance for `calculate_implied_rate` scale with the total implied rate, to ensure the tests don't fail while the error is still within reasonable bounds for larger results.

Adds `calculate_hpr_given_apr` and `calculate_hpr_given_apy` functions.

# Review Checklists

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-math/src/short/open.rs
+++ b/crates/hyperdrive-math/src/short/open.rs
@@ -758,14 +758,19 @@ mod tests {
 
             // Bob closes his short.
             let base_proceeds = bob.close_short(maturity_time, bond_amount, None).await?;
-            let position_duration = bob.get_state().await?.annualized_position_duration();
+            let annualized_position_duration =
+                bob.get_state().await?.annualized_position_duration();
 
             // Ensure that the implied rate matches the realized rate from
             // holding the short to maturity.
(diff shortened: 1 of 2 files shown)


crates/hyperdrive-math/src/short/open.rs
```diff
@@ -754,20 +758,27 @@ mod tests {
 
             // Bob closes his short.
             let base_proceeds = bob.close_short(maturity_time, bond_amount, None).await?;
+            let position_duration = bob.get_state().await?.annualized_position_duration();
```
dpaiton, line 761: I would call this `annualized_position_duration` to avoid confusion from the `position_duration` attribute.

dpaiton (approved):

crates/hyperdrive-math/src/short/open.rs
MazyGio: Updated, thanks!

Issue #99, opened 2024-05-15
Refactor to return errors instead of panic


dpaiton:
## TL;DR
Our use of both `panic` and `result` results in a worse developer experience, and should be changed to using `Result` everywhere.

## The problem
Per the [Rust docs](https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html#to-panic-or-not-to-panic):

> You could call panic! for any error situation, whether there’s a possible way to recover or not, but then you’re making the decision that a situation is unrecoverable on behalf of the calling code.

While it was reasonable to set up panics in the MVP stage of this repo, it has matured enough now that we should take care to consider developer experience with error handling. The current hybrid workflow uses `panic` _and_ `Result`, which is difficult to manage.

## Examples
We have stuff like this:
```rs
            let max_long = match panic::catch_unwind(|| {
                state.calculate_max_long(U256::MAX, checkpoint_exposure, None)
            }) {
                Ok(max_long) => match max_long {
                    Ok(max_long) => max_long,
                    Err(_) => continue,
                },
                Err(_) => continue,
            };
```

Where we have to catch both a panic and a result to handle a type `<Result, Result<FixedPoint, Report>, Report>`. 

Another situation is here:

```rs
            // Need to catch panics because of FixedPoint.
            let actual = panic::catch_unwind(|| {
                state.calculate_max_long(
                    U256::MAX,
                    checkpoint_exposure,
                    Some(max_iterations.into()),
                )
            });
            match chain
                .mock_hyperdrive_math()
                .calculate_max_long(
                    MaxTradeParams {
                        share_reserves: state.info.share_reserves,
                        bond_reserves: state.info.bond_reserves,
                        longs_outstanding: state.info.longs_outstanding,
                        long_exposure: state.info.long_exposure,
                        share_adjustment: state.info.share_adjustment,
                        time_stretch: state.config.time_stretch,
                        vault_share_price: state.info.vault_share_price,
                        initial_vault_share_price: state.config.initial_vault_share_price,
                        minimum_share_reserves: state.config.minimum_share_reserves,
                        curve_fee: state.config.fees.curve,
                        flat_fee: state.config.fees.flat,
                        governance_lp_fee: state.config.fees.governance_lp,
                    },
                    checkpoint_exposure,
                    max_iterations.into(),
                )
                .call()
                .await
            {
                Ok((expected_base_amount, ..)) => {
                    assert_eq!(
[...]

Issue #66, opened 2024-05-03
add rust lint ignore for match cmp


dpaiton:
Rust lint (clippy) suggests we make a change from 

```
if x < y
else if x > y
else
```

to

```
match x.cmp(y)
ordering.less
ordering.greater
ordering.equal
```

This is a pretty [controversial](https://github.com/rust-lang/rust-clippy/issues/4531) lint suggestion from clippy. @jalextowle  and I talked about it and we both agree that the original formatting is better (more readable, doesn't require extra import). We should set a global ignore for this lint warning.

Issue #65, opened 2024-05-03
fix type conversion error messaging to reflect FixedPoint conversion instead of U256


dpaiton:
In the python hyperdrive bindings we do a lot of unit conversions to handle communications between python and rust.

ultimately the thing we're trying to do is convert to fixedpoint. The way we do it everywhere is assume that the error will come from the string->U256 conversion and not the U256->FixedPoint conversion. Since we have the error attached to the string->U256 conversion we say that ("failed to convert to U256") in the error message. We should be more clear that we were trying to convert to fixed point and it failed in the U256->string conversion.

dpaiton:
I looked through this more closely and I decided to leave the errors alone. The error it self comes from the U256 conversion, so that is what it should say. I added the actual error message in #113 to make things more clear.

Issue #64, opened 2024-05-03
change `unwrap` to `map_err` in hyperdrive python bindings


dpaiton:
The best example is in `calculate_targeted_long_with_budget`, where we do:

```
            .map_err(|err| {
                PyErr::new::<PyValueError, _>(format!(
                    "Calculate_targeted_long_with_budget returned the error: {:?}",
                    err
                ))
            })?;
```

But that's the only place we do it.

In `calculate_spot_price_after_short` we do something like that but throw away the actual `err`. In most other fns we use `unwrap()`, which throws a cryptic panic message. We should follow the targeted long example everywhere.

PR #113, opened 2024-05-23
Replace panic with Result

dpaiton:
# Resolved Issues
https://github.com/delvtech/hyperdrive-rs/issues/99 https://github.com/delvtech/hyperdrive-rs/issues/64 https://github.com/delvtech/hyperdrive-rs/issues/65 https://github.com/delvtech/hyperdrive-rs/issues/66

# Description
Our use of both panic and result results in a worse developer experience, and should be changed to using Result everywhere possible. This PR removes almost all `panic` uses in the core libs; where functions once would panic but return a type `T` they now return an error with the type`Result<T>`. I also removed `unwrap()` statements, which convert `Result` to `panic`, in most places around the repo.

 Exceptions were made for core (anything with sugar like `+`, `-`, `*`) FixedPoint arithmetic (thus keeping any U256 panics), certain test-related cases, and macros.

I also improved some error messaging (in bindings mostly, but also elsewhere), added comments, & added a lint ignore.

I ran the fuzz testing with 10k `FUZZ_RUNS` and 50k `FAST_FUZZ_RUNS` twice without error locally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,3 +21,6 @@ authors = [
     "Ryan Goree <ryan@delv.tech>",
     "Sheng Lundquist <sheng@delv.tech>",
 ]
+
+[workspace.lints.clippy]
+comparison_chain = "allow"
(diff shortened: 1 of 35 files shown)


sentilesdal (approved):
Yeah this all seems fine to me.  Add some more in the description if you can.  Looks like you're:
  - getting rid of all explicit `panic!` calls in hyperdrive-math
  - fixedpoint pow returns result
  - fixedpoint exp returns result
  - error messaging improvements in bindings
  - catch overflow panics for max short/long

PR #118, opened 2024-05-31
Don't rebuild hyperdrive-wrappers if no solidity files changed.

sentilesdal:
# Resolved Issues
n/a

# Description
Building hyperdrive-wrappers takes a long time and happens way too often during local development and testing.  We don't need to rebuild these if no solidity files have changed.  On top of this, vscode is breaking because its watching for changes and rebuilding way too much.  This fixes both of these issues.

# Review Checklists

- [ ] **Testing**

### DIFF
--- a/crates/hyperdrive-wrappers/build.rs
+++ b/crates/hyperdrive-wrappers/build.rs
@@ -1,17 +1,19 @@
 use std::{
     collections::HashMap,
     env,
-    fs::{create_dir_all, read_to_string, OpenOptions},
+    fs::{self, create_dir_all, read_to_string, OpenOptions},
     io::Write,
-    path::Path,
+    path::{Path, PathBuf},
     process::Command,
 };
 
 use dotenv::dotenv;
 use ethers::prelude::Abigen;
 use eyre::Result;
+use filetime::FileTime;
 use heck::ToSnakeCase;
 use serde::{Deserialize, Serialize};
+use walkdir::WalkDir;
 
 const HYPERDRIVE_URL: &str = "https://github.com/delvtech/hyperdrive.git";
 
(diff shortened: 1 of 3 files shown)


crates/hyperdrive-wrappers/build.rs
```diff
@@ -1,17 +1,19 @@
 use std::{
     collections::HashMap,
     env,
-    fs::{create_dir_all, read_to_string, OpenOptions},
+    fs::{self, create_dir_all, read_to_string, OpenOptions},
```
dpaiton, line 4: are you sure you want to import fs::self here? that seems odd

dpaiton: is it because of `fs::read_dir`? could you just import `read_dir` and not self and then avoid the `fs::` later?

dpaiton (approved):

wakamex:
another way to speed this up is just to remove the `--force` from the `forge build` command, which already checks whether solidity files have changed:

```
forge build
[⠒] Compiling...
No files changed, compilation skipped
```

wakamex:
you could also check the output of the `forge build` to see whether the wrappers need to be rebuilt (whether or not `No files changed, compilation skipped` is in the output)

sentilesdal:
@wakamex that's brilliant.  Gonna try that out.

crates/hyperdrive-wrappers/build.rs
sentilesdal: yeah good call, fixed this.

crates/hyperdrive-wrappers/build.rs
```diff
@@ -67,56 +72,45 @@ const TARGETS: &[&str] = &[
...
-    // Compile the contracts.
+    // Compile the contracts and exit early if there are no changes.
     Command::new("forge")
         .current_dir(&hyperdrive_dir)
         .arg("install")
         .output()?;
     Command::new("forge")
         .current_dir(&hyperdrive_dir)
-        .args(["build", "--force"])
+        .args(["build"])
```
ryangoree, line 102: nit:
```suggestion
        .arg("build")
```

ryangoree (approved):

crates/hyperdrive-wrappers/build.rs
sentilesdal: gotcha

Issue #119, opened 2024-06-01
Implied rate (Short ROI) is negative in some cases where fixed > variable


MazyGio:
<img width="622" alt="_💅︱frontend___DELV__Fmr_Element_Finance__-_Discord" src="https://github.com/delvtech/hyperdrive-rs/assets/95626798/81dd7ae3-a84c-458b-b383-599dbe3e70ce">

The calc for implied rate doesn't currently account for backpaid interest. It's based on the full amount of basePaid, when we should be working with `basePaid - backpaidInterest`, since the user gets it back when closing the short, so it's not really a "cost":

`backpaidInterest = bondAmount * (vaultSharePrice/openVaultSharePrice) - bondAmount`

PR #120, opened 2024-06-01
fix calculate_implied_rate to exclude the backpaid interest from the cost basis

MazyGio:
# Resolved Issues
#119 Implied rate (Short ROI) is negative in some cases where fixed > variable

# Description
This excludes the short's backpaid interest from implied_rate calc

# Review Checklists

- [ ] **Testing**


dpaiton (approved):
This lgtm although I think it would be a good idea to put a test together that demonstrates how it works as a follow-up.

Issue #21, opened 2024-03-19
`calculate_targeted_short`


jalextowle:
We should add a function to the Rust SDK that calculates the short amount that brings the pool to a targeted effective rate (including fees). Traders could use this to calculate the optimal trade to make given their expected future variable rate.

wakamex:
here's a log from a crash in `bots_on_testnet`

```
random_steth30 (0x8817b7e94dc29503839f6515bc2652aBd80b2127) BASE=508 ETH=0.04967
===POOL INFO===
PoolInfoFP(share_reserves=FixedPoint("346643.960988065950165133"), share_adjustment=FixedPoint("7882.266532349756602536"), zombie_base_proceeds=FixedPoint("0.0"), zombie_share_reserves=FixedPoint("0.0"), bond_reserves=FixedPoint("29223865.18427352698382377"), lp_total_supply=FixedPoint("292689.997757872582173643"), vault_share_price=FixedPoint("1.00158206576483243"), longs_outstanding=FixedPoint("681775.196930321061294182"), long_average_maturity_time=FixedPoint("1716249179.401359641739904204"), shorts_outstanding=FixedPoint("627128.24131681"), short_average_maturity_time=FixedPoint("1716315144.100679191880729482"), withdrawal_shares_ready_to_withdraw=FixedPoint("0.000000000000000701"), withdrawal_shares_proceeds=FixedPoint("0.000000000000000701"), lp_share_price=FixedPoint("1.001191756023741379"), long_exposure=FixedPoint("305565.268200216168602777"))
Traceback (most recent call last):
  File "/code/agent0/scripts/bots_on_testnet.py", line 208, in <module>
    event_list = agent.execute_policy_action()
  File "/code/agent0/src/agent0/core/hyperdrive/interactive/i_hyperdrive_agent.py", line 200, in execute_policy_action
    return self._pool._execute_policy_action(self.agent)
  File "/code/agent0/src/agent0/core/hyperdrive/interactive/i_hyperdrive.py", line 359, in _execute_policy_action
    trade_results: list[TradeResult] = asyncio.run(
  File "/code/agent0/.venv/lib/python3.10/site-packages/nest_asyncio.py", line 30, in run
    return loop.run_until_complete(task)
  File "/code/agent0/.venv/lib/python3.10/site-packages/nest_asyncio.py", line 98, in run_until_complete
    return f.result()
  File "/usr/lib/python3.10/asyncio/futures.py", line 201, in result
[...]

PR #123, opened 2024-06-04
add `calculate_short_proceeds_up` 

dpaiton:
# Resolved Issues
Partially addresses https://github.com/delvtech/hyperdrive-rs/issues/29 and https://github.com/delvtech/hyperdrive-rs/issues/21

# Description
These are steps towards implementing a parity implementation of open short as well as targeted short.
- Add a function to calc short proceeds & round up, which is used in the Open Short flow.
- Modifies calculate short proceeds down to look more like the up version.

Since these functions are implemented on State, we preferentially use State member variables (e.g. `vault_share_price`) instead of passing arguments.

The short proceeds is given by:

$$
P_{\text{short}}(\Delta y) = \tfrac{\Delta y \cdot c_{1}}{c_{0} \cdot c} + \tfrac{\Delta y}{c} \cdot \phi_{f} - \Delta z(\Delta y)
$$

### DIFF
--- a/crates/hyperdrive-math/src/short/close.rs
+++ b/crates/hyperdrive-math/src/short/close.rs
@@ -56,28 +56,123 @@ impl State {
         Ok(flat + curve)
     }
 
-    // Calculates the proceeds in shares of closing a short position.
-    fn calculate_short_proceeds(
+    /// Calculates the proceeds in shares of closing a short position. This
+    /// takes into account the trading profits, the interest that was
+    /// earned by the short, the flat fee the short pays, and the amount of
+    /// margin that was released by closing the short. The math for the
+    /// short's proceeds in base is given by:
+    ///
+    /// $$
+    /// proceeds = (\frac{c1}{c_0 \cdot c}
+    /// + \text{\phi_f}) \cdot \frac{\Delta y}{c}
+    /// - dz
+    /// $$
+    ///
+    /// We convert the proceeds to shares by dividing by the current vault
+    /// share price. In the event that the interest is negative and
+    /// outweighs the trading profits and margin released, the short's
+    /// proceeds are marked to zero.
+    pub fn calculate_short_proceeds_up(
         &self,
         bond_amount: FixedPoint,
         share_amount: FixedPoint,
         open_vault_share_price: FixedPoint,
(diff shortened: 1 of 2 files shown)


crates/hyperdrive-math/src/short/close.rs
```diff
@@ -56,28 +56,123 @@ impl State {
...
+        } else {
+            fixed!(0)
+        }
+    }
+
+    /// Returns the derivative of the short proceeds calculation, assuming that the interest is
+    /// less negative than the trading profits and margin released.
+    ///
+    /// $$
+    /// P^{\prime}_{\text{short}}(\Delta y) = \tfrac{c_{1}}{c_{0} \cdot c}
```
jalextowle, line 116: Damn, this is some nice latex

crates/hyperdrive-math/src/short/close.rs
```diff
@@ -56,28 +56,123 @@ impl State {
...
+        // interest proceeds, and the margin released.
+        if total_value > share_amount {
+            // proceeds = (c1 / c0 * c) * dy - dz
+            total_value - share_amount
+        } else {
+            fixed!(0)
+        }
+    }
+
+    /// Returns the derivative of the short proceeds calculation, assuming that the interest is
```
jalextowle, line 112: I know it's obvious in context, but might be good to say "with respect to the bond amount"

crates/hyperdrive-math/src/short/close.rs
jalextowle, line 116: Isn't this missing the derivative of $\Delta z(\Delta y)$? The share reserves delta is a function in terms of the bond amount.

jalextowle: I was expecting to see something equal to the `short_deposit_derivative` but maybe I'm missing something.

dpaiton: oh interesting yeah, I was treading ∆z as a constant. You're right though. ∆z(∆y) is different for open & close, so I think I will need to have two different derivative calculations. Do you agree?

jalextowle (approved):

PR #134, opened 2024-06-09
Hyperdrive v1.0.10

dpaiton:
updates the registry & factory deployment to support new functions added in hyperdrive v1.0.10


crates/hyperdrive-test-utils/src/chain/deploy.rs
```diff
@@ -919,7 +922,6 @@ impl TestnetDeploy for Chain {
         base.transfer_ownership(config.admin).send().await?;
         vault.transfer_ownership(config.admin).send().await?;
         lido.transfer_ownership(config.admin).send().await?;
-        factory.update_governance(config.admin).send().await?;
```
jalextowle, line 922: I could be missing something, but I don't think we want to remove this. What was the impetus for this change?

crates/hyperdrive-test-utils/src/chain/deploy.rs
```diff
@@ -349,6 +348,7 @@ impl TestnetDeploy for Chain {
             },
         };
```
jalextowle, line 350: Nit: Let's remove this newline. It breaks up the flow of the code IMO

jalextowle (approved):

crates/hyperdrive-test-utils/src/chain/deploy.rs
```diff
@@ -543,6 +541,8 @@ impl TestnetDeploy for Chain {
             .await?
         };
 
+        let lp_math = LPMath::deploy(client.clone(), ())?.send().await?;
```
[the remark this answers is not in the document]
dpaiton: I did not add any lines here, I only moved it to be closer to where it is used.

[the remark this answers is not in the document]
dpaiton: looks like it is called in two different functions

crates/hyperdrive-test-utils/src/chain/deploy.rs
dpaiton: The impetus is that `UpdateGovernance` is gone now. and my understanding is we only have to do UpdateAdmin in the registry and not the factory.

jalextowle: I'm not sure what you mean. It's still there on main: https://github.com/delvtech/hyperdrive/blob/main/contracts/src/factory/HyperdriveFactory.sol#L354.

dpaiton: got it, I was mixing up where that was deleted. I added it back in.

Issue #136, opened 2024-06-11
Max long & short failures


dpaiton:
## Short

Fuzz testing is seeing the following errors:
- `calculate_max_long`
   - `InsufficientLiquidity: Negative Interest`
   - `Initial guess in calculate_max_long is insolvent`
   - `target share reserves less than effective share reserves`
- `calculate_max_short`
   - `absolute_max_short is insolvent`
   - `arithmetic operation overflow`

The fuzz pool is initialized to a known-good state (@slundqui is 80% certain it is good based on visual inspection of the parameters from himself, Jonny, & Alex and comparison against rust CI tests).
Random bots make trades and call `calc_max_long` and `calc_max_short` at random times throughout.

-------------------------------

`fuzz_calculate_max_short` is failing with large differences in budget amount: https://github.com/delvtech/hyperdrive-rs/issues/121
- This is happening rarely, but I believe that is because the test itself is rarely hitting the assert. I rewrote the test to always hit the assert and it fails much more often.

dpaiton:
Related to https://github.com/delvtech/hyperdrive-rs/issues/185

dpaiton:
- [ ] This line was recently added in #207 and should be removed once max long is fixed (because the max should always be solvent):

in `test_calculate_max_long`
```rs
// Ensure max long is valid.
if state.calculate_open_long(max_long).is_err() {
    continue;
}
```

- [ ] These test are being ignored and should be revisited  by whoever picks up this project:
`fuzz_sol_solvency_after_short`
`fuzz_calculate_max_short_budget_consumed`

- [ ] `fuzz_absolute_max_short_valid` does not pass 100% of the time due to pool state after preamble not supporting a valid short

Issue #121, opened 2024-06-02
`fuzz_calculate_max_short` occasionally failing


dpaiton:
```
---- short::max::tests::fuzz_calculate_max_short stdout ----
thread 'short::max::tests::fuzz_calculate_max_short' panicked at crates/hyperdrive-math/src/short/max.rs:742:17:
expected (base=70430895.800062968817706189) < (budget=70964720.871387510943216964) * 0.001000000000000000 = 70964.720871387510943216
```

I posted about a similar error [here](https://github.com/delvtech/hyperdrive-rs/issues/104#issuecomment-2138347965) but closed that issue in favor of this dedicated one.

wakamex:
my testing shows a problem occurring in a specific range of fixed rates from about 100% to 300%:

```
Total tests: 17720 Both pass: 8496 Both fail: 8873 Mismatched: 351 Failure rate: 0.019808126410835215
Fuzzed over fixed rate from Some(FixedPoint(0.000385747703397417)) to Some(FixedPoint(399.625149689078101990))
Mismatched  fixed rate from Some(FixedPoint(3.131040290023483491)) to Some(FixedPoint(133.805637502892130607))
```

that's from wip draft pr #19

wakamex:
still get a 2% mismatch rate after #122 allows the test to run to the end

```
Total tests: 100,000 Both pass: 48,054 Both fail: 49,871 Mismatched: 2,075 Failure rate: 0.02075
```

wakamex:
looks like Rust succeeds when Solidity doesn't:

```
Fixed rate: 9.258702584950471355
MISMATCHED: actual: Ok(Ok(FixedPoint(44767131.875921600404760400))) expected: Revert(Bytes(0x4e487b710000000000000000000000000000000000000000000000000000000000000011))
Rust: FixedPoint(44767131.875921600404760400) Solidity: Revert(Bytes(0x4e487b710000000000000000000000000000000000000000000000000000000000000011))
Fixed rate: 5.460357484762586677
MISMATCHED: actual: Ok(Ok(FixedPoint(835897164.703558754582240549))) expected: Revert(Bytes(0x4e487b710000000000000000000000000000000000000000000000000000000000000011))
```

this result is on 989ad42c89e4b9179331f8b879e376a356433bca

dpaiton:
Fuzz testing is seeing the following errors:
- `calculate_max_long`
   - `InsufficientLiquidity: Negative Interest`
[...]

PR #137, opened 2024-06-12
max short fixes

dpaiton:
# Resolved Issues
Partially addresses #121 & #136

# Description
These changes arose in my other pr (#116) while trying to debug issues and high tolerances in `short::max`.  There are a lot of commits in order to separate out meaningful changes from trivial ones. A majority of the changes are code reorganization, docstring fixes, and some additional sanity checks.

Major changes were:
- `fuzz_calculate_max_short`
  - was renamed to more accurately reflect what it is testing -- max short in the budget constrained regime
  - was rewritten to ensure the budget is constrained. I believe the reason the failures were "intermittent" before was because it was not always hitting the failure case. It hits it a lot now.
  - the tolerance was updated so that it consistently passes, to give us an idea of how much it is off from what we want.
  
- `calculate_max_short`
this code was incorrect:
 ```
          let absolute_max_deposit =
            match self.calculate_open_short(max_bond_amount, open_vault_share_price) {
                Ok(d) => d,
                Err(_) => return Ok(max_bond_amount),
``` 

we do **not** want to return the bond amount if `calculate_open_short` is throwing an error. I updated it.

I ran the tests locally with `FUZZ_RUNS=1_000` and `FAST_FUZZ_RUNS=50_000` without any tests failing.

### DIFF
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -70,7 +70,7 @@ impl State {
         let target_budget = if budget < self.minimum_transaction_amount().into() {
             return Ok(fixed!(0));
         }
-        // If the budget equals the minimum allowed, then we check if that's ok and then return.
+        // If the budget equals the minimum transaction amount, then we check if that's ok and then return.
         else if budget == self.minimum_transaction_amount().into() {
             if self
                 .solvency_after_short(
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -547,13 +547,13 @@ mod tests {
     /// `calculate_max_short` with a budget of `U256::MAX` to ensure that the two
     /// functions are equivalent.
     #[tokio::test]
-    async fn fuzz_sol_calculate_max_short_without_budget() -> Result<()> {
-        // TODO: We should be able to pass this with a much lower (if not zero) tolerance.
-        let tolerance = fixed!(1e17);
-
-        let chain = TestChain::new().await?;
+    async fn fuzz_sol_calculte_max_short_without_budget() -> Result<()> {
+        // TODO: We should be able to pass these tests with a much lower (if not zero) tolerance.
+        let sol_correctness_tolerance = fixed!(1e17);
+        let reserves_drained_tolerance = fixed!(1e28);
 
         // Fuzz the rust and solidity implementations against each other.
+        let chain = TestChain::new().await?;
         let mut rng = thread_rng();
         for _ in 0..*FAST_FUZZ_RUNS {
             let state = rng.gen::<State>();
@@ -601,6 +601,7 @@ mod tests {
                 .await
             {
                 Ok(expected) => {
+                    // Make sure the solidity & rust runctions gave the same value.
                     let actual_unwrapped = actual.unwrap().unwrap();
                     let expected_fp = FixedPoint::from(expected);
                     let error = if expected_fp > actual_unwrapped {
@@ -609,10 +610,27 @@ mod tests {
                         actual_unwrapped - expected_fp
                     };
                     assert!(
-                        error < tolerance,
+                        error < sol_correctness_tolerance,
                         "error {} exceeds tolerance of {}",
                         error,
-                        tolerance,
+                        sol_correctness_tolerance,
+                    );
+
+                    // Make sure the pool was drained.
+                    let effective_shares = state.effective_share_reserves()?;
+                    let min_shares = state.minimum_share_reserves();
+                    assert!(effective_shares >= min_shares,
+                        "effective_share_reserves={} should always be greater than the minimum_share_reserves={}.",
+                        state.effective_share_reserves()?,
+                        state.minimum_share_reserves()
+                    );
+                    let reserve_amount_above_minimum = effective_shares - min_shares;
+                    assert!(reserve_amount_above_minimum < reserves_drained_tolerance,
+                        "effective_share_reserves={} - minimum_share_reserves={} (diff={}) should be < tolerance={}",
+                        effective_shares,
+                        min_shares,
+                        reserve_amount_above_minimum,
+                        reserves_drained_tolerance,
                     );
                 }
                 Err(_) => {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: crates/hyperdrive-math/src/short/max.rs [hunk 1]
```diff
@@ -61,15 +62,34 @@ impl State {
...
+                    self.minimum_transaction_amount().into(),
+                    checkpoint_exposure,
+                )?
+                .is_some()
+            {
+                return Ok(self.minimum_transaction_amount().into());
+            }
+            return Ok(fixed!(0));
+        }
+        // If it's greater than the minimum allowed then we set the target budget.
```
jalextowle, line 86: Nit: 
```suggestion
        // If the budget is greater than the minimum transaction amount, then we set the target budget.
```

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -61,15 +62,34 @@ impl State {
...
-        if budget == fixed!(0) {
+        // To avoid the case where Newton's method overshoots and stays on
+        // the invalid side of the optimization equation (i.e., when deposit > budget),
+        // we artificially set the target budget to be less than the actual budget.
+        //
+        // If the budget is less than the minimum allowed, then we return early.
+        let target_budget = if budget < self.minimum_transaction_amount().into() {
             return Ok(fixed!(0));
         }
+        // If the budget equals the minimum allowed, then we check if that's ok and then return.
```
jalextowle, line 73: Nit:

```suggestion
        // If the budget equals the minimum transaction amount, then we check if that's ok and then return.
```

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -519,14 +548,17 @@ mod tests {
     /// functions are equivalent.
     #[tokio::test]
     async fn fuzz_calculate_max_short_no_budget() -> Result<()> {
```
jalextowle, line 550: This name could be improved IMO. It's not obvious that this is a differential test and the other isn't a differential test.

jalextowle: A related concern that I have with this test is that it's not verifying the invariants that we verify in the other test. I like the change that you made to the other test, but there should still be a test that verifies that our invariants hold with cases that aren't budget constrained.

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -717,37 +747,43 @@ mod tests {
...
-                    bob.base(),
-                    budget,
-                    error_tolerance,
-                    budget * error_tolerance
-                );
-            }
+            // Bob used a slippage tolerance of 0.01%, which means
+            // that the max short is always consuming at least 99.99% of
+            // the budget.
+            let budget_tolerance = fixed!(1e18); // TODO: This should be fixed!(0.0001e18) == 0.01%
```
jalextowle, line 774: Does this still need to be `1`? It seems like this makes the assertion useless, but I could be missing something.

dpaiton: any order of magnitude less than one causes it to fail unfortunately. This is (probably) the same issue as we were seeing here: https://github.com/delvtech/hyperdrive-rs/issues/121 but we weren't seeing it as often because the budget & max were not always lined up to create a condition where the assert was called. Now that I changed up the budget it gets called & breaks every time if the `budget_tolerance` is less than `1e18` (only checking orders of magnitude)

crates/hyperdrive-math/src/short/max.rs
dpaiton: > This name could be improved IMO. It's not obvious that this is a differential test and the other isn't a differential test.

agreed. I've renamed a bunch of tests to try to have consistency with `fuzz` == "runs executions with random inputs in a loop over `FUZZ_RUNS` or `FAST_FUZZ_RUNS`". vs `test` == "An isolated test run once". I think it makes sense to add a third name, `diff` or `comp` or `sol`, perhaps, that we can add to `fuzz`.

If you have a preference lets go with it, I don't. If you don't then I'll pick one and follow up with a mass rename in another PR.

dpaiton: > A related concern that I have with this test is that it's not verifying the invariants that we verify in the other test. I like the change that you made to the other test, but there should still be a test that verifies that our invariants hold with cases that aren't budget constrained.

Sure, what invariant checks should I add?

jalextowle: I like `diff` or `sol`, but it's totally up to you

crates/hyperdrive-math/src/short/max.rs
jalextowle: Ok gotcha. Does this mean that there are times where almost none of the budget is consumed? I'm wondering what the share reserves and effective share reserves look like in that case.

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -569,14 +601,36 @@ mod tests {
...
+                        state.minimum_share_reserves()
+                    );
+                    let reserve_amount_above_minimum = effective_shares - min_shares;
+                    assert!(reserve_amount_above_minimum < reserves_drained_tolerance,
+                        "effective_share_reserves={} - minimum_share_reserves={} (diff={}) should be < tolerance={}",
+                        effective_shares,
+                        min_shares,
+                        reserve_amount_above_minimum,
+                        reserves_drained_tolerance,
                     );
```
jalextowle, line 634: Ok, I think that one reason we may be having problems here is if the `share_reserves < effective_share_reserves` then the `share_reserves` will be equal to the minimum share reserves, but the effective share reserves will be greater (potentially by a lot). Since this test fuzzes over general pool states, I wouldn't be surprised if this is causing the issue.

We should be able to lower the tolerance if we check the minimum share reserves against `min(share_reserves, effective_share_reserves)`.

crates/hyperdrive-math/src/short/max.rs
dpaiton: went with sol

crates/hyperdrive-math/src/short/max.rs
dpaiton: That did not result in an improved tolerance, unfortunately. I am going to merge, with the idea that we can focus more on understanding what's going on in my next PR fixing open short.

crates/hyperdrive-math/src/short/max.rs
dpaiton: I agree something seems off. I'm going to merge this PR and see how things look after #116 is updated. That way we can rule out any of the fixes I already implemented there.

jalextowle (approved):
LGTM

PR #142, opened 2024-06-18
Dry up test preamble; code cleanup

dpaiton:
# Resolved Issues
Making testing more consistent to help narrow problems found when investigating https://github.com/delvtech/hyperdrive-rs/issues/29

# Description
- single preamble
We had 3x duplicates of a `preamble` function that would execute a series of trades on the deployed test pool in order to simulate testing from a random state. I consolidated it into one location that is imported elsewhere. I also added some checks on upper trade bounds so that it is guaranteed to be valid, even if calc_max fails.

- code cleanup
This seemingly innocuous change caused a difficult-to-track Hisenbug where the ERC20 contract would throw an overflow/underflow error whenever we'd call `hyperdrive.open_long` or `hyperdrive.open_short`. This was due to lack of approval, which was due to the approval call not going through before the open call, which was due to anvil not being able to keep up with the rust calls coming in. 
I fixed that bug (thanks to @mcclurejt for the help!) and in the process of hunting down that bug did a bunch of code cleanup stuff like improved error messaging, variable names, organization, etc.
I'm was able to consistently reproduce #4 while tracking this bug, and the fix we implemented seems to have also fixed that one.

- new tests
I also wrote a couple of new tests while hunting the Heisenbug.

### NOTE:
I had to increase test tolerance for `fuzz_sol_calc_open_short` from 1e9 to 10e18. My best guess as to why this happened is because we're now testing a wider variety of valid states and we're hitting error modes. I know that `calculate_open_short` does not match solidity exactly, so most likely this is the source of the error. I added a comment as a reminder in #29, but for now I think we should live with the high tolerance.

That being said, I encourage the reviewer to look carefully. I may have missed a change in this PR that is causing such a big error.

### DIFF
--- a/crates/hyperdrive-math/src/long/open.rs
+++ b/crates/hyperdrive-math/src/long/open.rs
@@ -132,67 +132,9 @@ mod tests {
     use rand_chacha::ChaCha8Rng;
 
     use super::*;
-    use crate::test_utils::agent::HyperdriveMathAgent;
-
-    /// Executes random trades throughout a Hyperdrive term.
-    async fn preamble(
-        rng: &mut ChaCha8Rng,
-        alice: &mut Agent<ChainClient<LocalWallet>, ChaCha8Rng>,
-        bob: &mut Agent<ChainClient<LocalWallet>, ChaCha8Rng>,
-        celine: &mut Agent<ChainClient<LocalWallet>, ChaCha8Rng>,
-        fixed_rate: FixedPoint,
-    ) -> Result<()> {
-        // Fund the agent accounts and initialize the pool.
-        alice
-            .fund(rng.gen_range(fixed!(1_000e18)..=fixed!(500_000_000e18)))
-            .await?;
-        bob.fund(rng.gen_range(fixed!(1_000e18)..=fixed!(500_000_000e18)))
-            .await?;
-        celine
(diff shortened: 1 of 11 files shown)


crates/hyperdrive-math/src/long/open.rs
```diff
@@ -357,23 +350,23 @@ mod tests {
...
-                * FixedPoint::from(state.info.vault_share_price);
-            let long_amount = rng.gen_range(min_base_amount..=max_long);
+            let base_amount = rng.gen_range(min_txn_amount..=max_long);
 
             // Compare the open short call output against calculate_open_long.
-            let bonds_purchased = state.calculate_open_long(long_amount);
+            let rust_bonds = state.calculate_open_long(base_amount);
 
+            // Fund a little extra to allow for of slippage.
+            bob.fund(base_amount + fixed!(10e18)).await?;
```
mcclurejt, line 365: What do you think of funding extra by a percentage of the `base_amount` as opposed to a flat value?

mcclurejt (approved):
LGTM! Left a minor suggestion

crates/hyperdrive-math/src/long/open.rs
dpaiton: Good idea! I'll add it.

PR #145, opened 2024-06-18
support hyperdrive 1.0.12

dpaiton:
Updates the hyperdrive dependency to version `1.0.12` & makes required rust changes.

A test is failing with this update, so I am ignoring it for now. It will pass once https://github.com/delvtech/hyperdrive-rs/pull/143 lands.


bindings/hyperdrivepy/python/hyperdrivepy/pypechain_types/IHyperdriveContract.py
```diff
@@ -636,6 +636,33 @@ def call(
         return cast(bool, rename_returned_types(structs, return_types, raw_values))
 
 
+class IHyperdriveKindContractFunction(ContractFunction):
```
jalextowle, line 639: This is generated by pypechain, right? If so, what is the benefit of committing these files? For contract wrappers in other languages, we gitignore wrappers since they are built during or before compilation

jalextowle (approved):

bindings/hyperdrivepy/python/hyperdrivepy/pypechain_types/IHyperdriveContract.py
dpaiton: Yes, they're generated by pypechain. I am not sure who made that call (could have been me) and I don't recall why. I think what you're saying makes sense, though. I will look into it when I have a chance.

Issue #141, opened 2024-06-17
`build.rs` does not always work


dpaiton:
on fresh installs we need to add a `forge install` step so that the Hyperdrive dependencies (openzeppelin, etc) are installed. When running make build on a fresh ubuntu environment it fails because this step is missing.

the `need_to_build` function also does not always work correctly. When doing a `LOCAL_DEVELOPMENT=true` build I was frequently getting unresolved import errors:
<img width="727" alt="image" src="https://github.com/delvtech/hyperdrive-rs/assets/153468/8e5b0008-b6e3-4a37-90b0-a9f820342086">

which are happening because it's not building. I could consistently fix the problem by forcing `need_to_build` to always return `Ok(true)`.

dpaiton:
the [--root](https://book.getfoundry.sh/reference/forge/forge-install#:~:text=Project%20Options-,%2D%2Droot%20path,-The%20project%E2%80%99s%20root) flag should be helpful in the script (`forge install --root path/to/hyperdrive`)

PR #148, opened 2024-06-21
manually disable build

sentilesdal:
# Resolved Issues
#141 

# Description
Instead of automatically checking if files have changed, just make this process manual via the `BUILD_DISABLED` environment variable.  Most of the time for development this can just be set to `true` if solidity files aren't being worked on.  If solidity files are being edited then the variable can be set to `false` to enable automatic rebuilding.

# Review Checklists

- [ ] **Testing**


dpaiton (approved):
Can you add this & the other build flags & `HYPERDRIVE_ETHEREUM_URL` to the README?

PR #151, opened 2024-06-22
short code cleanup

dpaiton:
# Resolved Issues
Working towards #29 

# Description
This pulls a bunch of cleanup changes out of #116 to be reviewed independently.
- fix a bug introduced in https://github.com/delvtech/hyperdrive-rs/pull/142 where test tolerances were high because the variable rate was causing short amounts to change between the rust & sol calls. 
- improve docstrings & error messaging throughout
- modify `solvency_after_short_derivative` to return `Result<T>` instead of `Result<Option<T>>`
- modify max behavior to throw errors if there is no valid max, instead of returning `0`
- add safety bounds on max short guesses to ensure it is always >= the min txn amount
- modify sol differential max short tests to use `calculate_absolute_max_short` instead of `calculate_max_short` since the solidity implementation does not consider budget
- removed the `fuzz_calculate_absolute_max_short_execute` in favor of a test that does the same but additionally checks that the pool is drained when that absolute max short is executed. It also includes some differential checks for rust vs solidity. This one is now called `fuzz_calculate_max_short_without_budget_then_open_short `
- adds a new `SLOW_FUZZ_RUNS` constant for one of the max tests bc it was slow as hell.

### DIFF
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -194,8 +194,8 @@ impl State {
             // Iteratively update max_bond_amount via Newton's method.
             let derivative = self.calculate_open_short_derivative(
                 max_bond_amount,
-                spot_price,
                 open_vault_share_price,
+                Some(spot_price),
             )?;
             if deposit < target_budget {
                 max_bond_amount += (target_budget - deposit) / derivative
@@ -754,7 +754,8 @@ mod tests {
 
     #[tokio::test]
     async fn fuzz_sol_calculate_max_short_without_budget_then_open_short() -> Result<()> {
-        let same_max_base_tolerance = fixed!(10e18);
+        let max_base_tolerance = fixed!(10e18);
+        let max_bonds_tolerance = fixed!(1e2);
         let reserves_drained_tolerance = fixed!(1e27);
 
         // Set up a random number generator. We use ChaCha8Rng with a randomly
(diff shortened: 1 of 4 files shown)


crates/hyperdrive-math/src/short/max.rs
```diff
@@ -100,18 +105,21 @@ impl State {
...
-        if absolute_max_bond_amount < self.minimum_transaction_amount().into() {
-            return Ok(fixed!(0));
+        if absolute_max_bond_amount < self.minimum_transaction_amount() {
+            return Err(eyre!("No solvent short is possible."));
         }
+
         // Figure out the base deposit for the absolute max bond amount.
         // This could fail due to inaccuracies in the absolute max estimate.
         // If so, ignore it and move on.
+        // FIXME: This should not fail; we should be able to use ? operator and move on.
```
dpaiton, line 122: I'll address this in a follow-up PR

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -432,12 +435,13 @@ impl State {
     ) -> Result<FixedPoint> {
         let checkpoint_exposure_shares =
             FixedPoint::try_from(checkpoint_exposure.max(I256::zero()))? / self.vault_share_price();
+        // FIXME: Check this against solidity; should we be using solvency here or just share reserves?
```
dpaiton, line 438: I'll address this in a follow-up PR

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -864,26 +864,31 @@ mod tests {
             // Run the preamble.
             initialize_pool_with_random_state(&mut rng, &mut alice, &mut bob, &mut celine).await?;
 
+            // Set the variable rate to 0.
+            // This is required so that no interest is accrued between the
+            // estimate and actual open_short call.
+            alice.advance_time(fixed!(0), fixed!(1)).await?;
```
sentilesdal, line 870: this doesn't make sense, something else must be wrong.  this just makes sure that the vault share price isn't 1.  the short is estimated/opened after this so they should still be the same no?

sentilesdal (commented):
shouldnt need to set variable to zero when advancng time.

sentilesdal (approved):
Just add a comment that setting the variable rate to zero is necessary to get the test to pass and that this is very bad and needs to be fixed

crates/hyperdrive-math/src/short/open.rs
dpaiton: 👍 added a comment per our discussion

Issue #27, opened 2024-04-16
Auto-generate Rust Documentation


jalextowle:
We should publish the auto-generated rust docs from the Rust SDK. We could also publish a docs portal for the Rust SDK and other SDKs.

dpaiton:
whoever dives into this should look into latex rendering (e.g. via [ktex_doc](https://docs.rs/katex-doc/0.1.0/katex_doc/)). If we are able to do it without too much friction, then I propose we use latex style docs for function (& above) docstrings, and human-readable docs for inline equations.

dpaiton:
All solutions that I've found are variations of using the `--html-in-header` flag to inject html that runs [katex](https://www.npmjs.com/package/katex) javascript at runtime.

I don't think we need this now, but one interesting find is that you can also use [`--html-before-content`](https://github.com/rust-lang/rust/issues/34261#issuecomment-1042696542) to define global LaTeX equation macros that then can be used throughout. 

Here's an example repo the renders latex docs: https://docs.rs/levenberg-marquardt/latest/levenberg_marquardt/

dpaiton:
ok, I think I got a handle on it. I'll take lead on this
<img width="1022" alt="image" src="https://github.com/delvtech/hyperdrive-rs/assets/153468/36a814c4-df23-4ea4-9e6a-90944c62dd43">

PR #152, opened 2024-06-23
update `calc_open_short` to match solidity

dpaiton:
# Resolved Issues
https://github.com/delvtech/hyperdrive-rs/issues/29
https://github.com/delvtech/hyperdrive-rs/issues/121

# Description
This PR updates the Rust open short and derivative implementations to match Solidity.
With help from @sentilesdal I uncovered a bug with sol differential tests where the current vault share price was not correctly used on the rust end. Fixing this allowed us to drop several test tolerances as well. I'll follow up with another PR that applies this fix elsewhere.
I also did some minor variable renames, added & modified max short checks, and resolved a few lingering FIXMEs.
I ran all of the tests with 10x fuzz constants locally without any failures.

## Lower tolerances on tests 🎉 
- After the fix, I was able to completely remove the error tolerance on `fuzz_calc_open_short`.
- I also updated the epsilon & tolerance values for the `fuzz_short_deposit_derivative` and `fuzz_short_principal_derivative` tests. I use a larger epsilon and smaller tolerance, which I think is what we prefer (more distant points should match linearity assumptions better due to the lack of monotonicity coming from `pow`).
- `fuzz_sol_calculate_max_short_without_budget_then_open_short` lowered base test tolerance due to both the open short fix and the vault share price fix.

## Increased tolerances on tests 😭 
- I had to increase tolerance on `fuzz_calculate_spot_price_after_short` from `fixed!(0)` to `fixed!(100)`. However, the test is more difficult now because it uses a random variable rate instead of 0%.
  - note: `test_sol_calculate_pool_deltas_after_open_short` looks like it has a new tolerance, but this was simply assigning an existing tolerance to a variable at the top of the test.
- `fuzz_sol_calculate_max_short_without_budget_then_open_short`  raised bond test tolerance from 1e2 to 1e10, but 1e2 wasn't holding up to extensive fuzz testing in the first place.

## Math outline
The below formulation follows the logic outlined in `HyperdriveShort.sol`. I made some slight changes to variable names and ordering of presentation.

image
image
image

### DIFF
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -117,17 +117,9 @@ impl State {
         }
 
         // Figure out the base deposit for the absolute max bond amount.
-        // This could fail due to inaccuracies in the absolute max estimate.
-        // If so, ignore it and move on.
-        // FIXME: This should not fail; we should be able to use ? operator and move on.
-        let maybe_absolute_max_deposit =
-            match self.calculate_open_short(absolute_max_bond_amount, open_vault_share_price) {
-                Ok(deposit) => Some(deposit),
-                Err(_) => None,
-            };
-        if maybe_absolute_max_deposit.is_some()
-            && maybe_absolute_max_deposit.unwrap() <= target_budget
-        {
+        let absolute_max_deposit =
+            self.calculate_open_short(absolute_max_bond_amount, open_vault_share_price)?;
+        if absolute_max_deposit <= target_budget {
             return Ok(absolute_max_bond_amount);
         }
 
@@ -433,13 +425,14 @@ impl State {
     ) -> Result<FixedPoint> {
         let checkpoint_exposure_shares =
             FixedPoint::try_from(checkpoint_exposure.max(I256::zero()))? / self.vault_share_price();
-        // FIXME: Check this against solidity; should we be using solvency here or just share reserves?
-        // solvency = share_reserves - long_exposure / vault_share_price - min_share_reserves
-        let solvency = self.calculate_solvency();
-        let guess = self.vault_share_price() * (solvency + checkpoint_exposure_shares);
+        let guess = self.vault_share_price().mul_down(
+            self.share_reserves() + checkpoint_exposure_shares
+                - self.long_exposure().div_down(self.vault_share_price())
+                - self.minimum_share_reserves(),
+        );
         let curve_fee = self.curve_fee() * (fixed!(1e18) - spot_price);
         let gov_curve_fee = self.governance_lp_fee() * curve_fee;
-        Ok(guess / (spot_price - curve_fee + gov_curve_fee))
+        Ok(guess.div_down(spot_price - curve_fee + gov_curve_fee))
     }
 
     /// Calculates the pool's solvency after opening a short.
@@ -888,18 +881,18 @@ mod tests {
                                 .min(state.share_reserves());
                             let min_share_reserves = state.minimum_share_reserves();
                             assert!(pool_shares >= min_share_reserves,
-                                    "effective_share_reserves={} should always be greater than the minimum_share_reserves={}.",
-                                    state.effective_share_reserves()?,
-                                    min_share_reserves,
-                                );
+                                "effective_share_reserves={} should always be greater than the minimum_share_reserves={}.",
+                                state.effective_share_reserves()?,
+                                min_share_reserves,
+                            );
                             let reserve_amount_above_minimum = pool_shares - min_share_reserves;
                             assert!(reserve_amount_above_minimum < reserves_drained_tolerance,
-                                    "share_reserves={} - minimum_share_reserves={} (diff={}) should be < tolerance={}",
-                                    pool_shares,
-                                    min_share_reserves,
-                                    reserve_amount_above_minimum,
-                                    reserves_drained_tolerance,
-                                );
+                                "share_reserves={} - minimum_share_reserves={} (diff={}) should be < tolerance={}",
+                                pool_shares,
+                                min_share_reserves,
+                                reserve_amount_above_minimum,
+                                reserves_drained_tolerance,
+                            );
                         }
 
                         // Solidity calculate_max_short worked, but passing that bond amount to open_short failed.
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -747,6 +747,8 @@ mod tests {
 
     #[tokio::test]
     async fn fuzz_sol_calculate_max_short_without_budget_then_open_short() -> Result<()> {
+        let max_bonds_tolerance = fixed!(1e10);
+        let max_base_tolerance = fixed!(1e10);
         let reserves_drained_tolerance = fixed!(1e27);
 
         // Set up a random number generator. We use ChaCha8Rng with a randomly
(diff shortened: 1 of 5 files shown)


crates/hyperdrive-math/src/short/max.rs
```diff
@@ -433,13 +425,14 @@ impl State {
...
         let checkpoint_exposure_shares =
             FixedPoint::try_from(checkpoint_exposure.max(I256::zero()))? / self.vault_share_price();
-        // FIXME: Check this against solidity; should we be using solvency here or just share reserves?
-        // solvency = share_reserves - long_exposure / vault_share_price - min_share_reserves
-        let solvency = self.calculate_solvency();
-        let guess = self.vault_share_price() * (solvency + checkpoint_exposure_shares);
+        let guess = self.vault_share_price().mul_down(
+            self.share_reserves() + checkpoint_exposure_shares
+                - self.long_exposure().div_down(self.vault_share_price())
+                - self.minimum_share_reserves(),
```
jalextowle, line 431: Did you update this so that it's more obvious what the components are? `calculate_solvency` should just be `self.share_reserves() - self.long_exposure().div_down(self.vault_share_price()) - self.minimum_share_reserves()` but maybe I'm missing something.

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -433,13 +425,14 @@ impl State {
...
+        );
         let curve_fee = self.curve_fee() * (fixed!(1e18) - spot_price);
         let gov_curve_fee = self.governance_lp_fee() * curve_fee;
-        Ok(guess / (spot_price - curve_fee + gov_curve_fee))
+        Ok(guess.div_down(spot_price - curve_fee + gov_curve_fee))
```
jalextowle, line 435: This is fine (and good in a lot of ways), but just to make sure that I understand, is the idea that you are trying to eliminate "magic" so that it's easier to quickly parse the rounding direction?

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -754,8 +747,8 @@ mod tests {
 
     #[tokio::test]
     async fn fuzz_sol_calculate_max_short_without_budget_then_open_short() -> Result<()> {
-        let max_base_tolerance = fixed!(10e18);
-        let max_bonds_tolerance = fixed!(1e2);
+        let max_bonds_tolerance = fixed!(1e10);
```
jalextowle, line 750: Why did the bond tolerance go up?

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -44,117 +70,197 @@ impl State {
...
+        // share reserves.
+        let share_reserves_delta = self.calculate_short_principal(bond_amount)?;
+
+        // NOTE: Round up to make the check stricter.
+        //
         // If the base proceeds of selling the bonds is greater than the bond
         // amount, then the trade occurred in the negative interest domain.
         // We revert in these pathological cases.
-        if share_reserves_delta_in_base > bond_amount {
+        if share_reserves_delta.mul_up(self.vault_share_price()) > bond_amount {
```
jalextowle, line 82: Great catch.

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -44,117 +70,197 @@ impl State {
...
+        // NOTE: Round up to overestimate the base deposit.
+        //
+        // The trader will need to deposit capital to pay for the fixed rate,
+        // the fees, and any back-paid interest that will be received back upon
+        // closing the trade.
+        let curve_fee_shares = self
+            .open_short_curve_fee(bond_amount)?
+            .div_up(self.vault_share_price());
+
+        if share_reserves_delta < curve_fee_shares {
```
jalextowle, line 100: Nit: Add a comment or get rid of the newline.

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -44,117 +70,197 @@ impl State {
...
+
+        // Now we can calculate adjusted proceeds account for the backdated
+        // vault price:
+        //
+        // $$
+        // \text{base_proceeds} = (
+        //    \frac{c1 \cdot \Delta y}{c0 \cdot c}
+        //    + \frac{\Delta y \cdot \phi_f}{c} - \Delta z
+        // ) \cdot c
+        // $$
```
jalextowle, line 124: Why are we using latex for these comments? It isn't going to get rendered, so it feels like it's just making it harder to read.

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -44,117 +70,197 @@ impl State {
...
         open_vault_share_price: FixedPoint,
         maybe_initial_spot_price: Option<FixedPoint>,
     ) -> Result<FixedPoint> {
+        // We're avoiding negative interest, so we will cap the close share
+        // price to be greater than or equal to the open share price. This
+        // will assume the max loss for the trader, some of which may be
+        // reimbursed upon closing.
+        let close_vault_share_price = open_vault_share_price.max(self.vault_share_price());
+
+        // Short circuit the derivative if the forward function returns 0.
```
jalextowle, line 182: Nit: Mind if we stick to "derivative" and "function"? Forward and backward seem a little out of place here, but it's not a hill that I'm going to die on.

crates/hyperdrive-math/src/short/open.rs
```diff
@@ -44,117 +70,197 @@ impl State {
...
-                - (self.bond_reserves() + bond_amount).pow(fixed!(1e18) - self.time_stretch())?);
-        // NOTE: The order of additions and subtractions is important to avoid underflows.
-        let payment_factor = (fixed!(1e18)
-            / (self.bond_reserves() + bond_amount).pow(self.time_stretch())?)
-            * theta.pow(self.time_stretch() / (fixed!(1e18) - self.time_stretch()))?;
-        Ok((self.vault_share_price() / open_vault_share_price)
-            + self.flat_fee()
-            + self.curve_fee() * (fixed!(1e18) - spot_price)
-            - payment_factor)
+        // All of these are in base
```
jalextowle, line 201: Nit: 
```suggestion
        // All of these are in base.
```

jalextowle (approved):
LGTM

crates/hyperdrive-math/src/short/max.rs
dpaiton: I was doing extensive fuzz testing on the `main` branch and found that `1e2` was not sufficient. So the bump is not directly due to this PR.

dpaiton: That being said, I think this is probably worth investigating further to see if it is a problem with the test or a problem with the code.

crates/hyperdrive-math/src/short/max.rs
dpaiton: I have been following a philosophy of functions that are supposed to mirror solidity should use that syntax (`div_down`) to make it easy for pattern matching and otherwise should use the more readable sugar (`/`).

This function happens to be a mix of both :-P I'll fix it.

crates/hyperdrive-math/src/short/open.rs
dpaiton: I'm going to leave this for now. I added a comment about it here https://github.com/delvtech/hyperdrive-rs/issues/27

crates/hyperdrive-math/src/short/open.rs
dpaiton: yes definitely -- I agree that phrasing should be reserved for ML code. Including it here was autonomic

crates/hyperdrive-math/src/short/max.rs
dpaiton: I double checked because I had forgotten the motivation here. The calculation was equivalent, and in general I am more of a fan of centralizing functions like `calculate_solvency`. I made this change to make it easier to directly compare to the solidity function `calculateMaxShortGuess`.

Issue #138, opened 2024-06-12
create functions for valuing positions using spot methodology


wakamex:
Similar to the frontend's use (https://github.com/delvtech/hyperdrive-frontend/issues/1144), we need spot methodology valuation in the rust sdk, which well then be using for ChainSync.

Previous discussion of spot versus closeout methodology is here: https://github.com/delvtech/hyperdrive/issues/1032.

I propose calling these functions `value_long` and `value_short` which to me makes it clear these don't include market impact of closing your account. Is this clear enough that this isn't the amount of money you'll get when you go to close your position? Would it be clearer if we called them something else? `market_value_long` and `market_value_short`? Looks like frontend is using `calcShortMarketValue` in https://github.com/delvtech/hyperdrive-frontend/pull/1167. For me the most clear would be `calc_long_closeout` and `calc_long_value`.

In keeping with the market standard of other trading platforms, would it make more sense not to include fees?

Plan:
- make these stateful (in long/close.rs and short/close.rs)
- make them as  close as possible to the existing closeout valuation functions (`calc_close_long`, `calc_close_short`)

formula for `value_long`:
```
market_estimate = trading_proceeds - flat_fees_paid - curve_fees_paid

trading_proceeds = flat_value + curve_value
flat_value       = dy * (1 - t)
curve_bonds      = dy * t
curve_value      = curve_bonds * p
                 = dy * t * p
flat_fees_paid   = flat_value * flat_fee
                 = dy * (1 - t) * flat_fee
curve_fees_paid  = (curve_bonds - curve_value) * curve_fee 
                 = dy * t * (1 - p) * curve_fee

dy = bond amount
p  = spotPrice
t  = termRemaining (positionDuration - (currentTimestamp - openTimestamp))
```

formula for `value_short`:
```
market_estimate = yield_accrued + trading_proceeds - curve_fees_paid + flat_fees_returned

yield_accrued      = dy * (c-c0)/c0
trading_proceeds   = dy * (1 - p) * t
curve_fees_paid    = trading_proceeds * curve_fee
flat_fees_returned = dy * t * flat_fee

dy = bond amount
c  = vaultSharePrice (current if non-matured, or checkpoint's if matured)
c0 = openVaultSharePrice
p  = spotPrice
t  = termRemaining (positionDuration - (currentTimestamp - openTimestamp))
```

PR #153, opened 2024-06-25
add market valuation functions

MazyGio:
# Resolved Issues
Adds `calculate_value_long` and `calculate_value_short` functions as described in #138 (https://github.com/delvtech/hyperdrive-rs/issues/138)

# Description
formula for `value_long`:
```
market_estimate = trading_proceeds - flat_fees_paid - curve_fees_paid

trading_proceeds = flat_value + curve_value
flat_value       = dy * (1 - t)
curve_bonds      = dy * t
curve_value      = curve_bonds * p
                 = dy * t * p
flat_fees_paid   = flat_value * flat_fee
                 = dy * (1 - t) * flat_fee
curve_fees_paid  = (curve_bonds - curve_value) * curve_fee 
                 = dy * t * (1 - p) * curve_fee

dy = bond amount
p  = spotPrice
t  = termRemaining (positionDuration - (currentTimestamp - openTimestamp))
```

formula for `value_short`:
```
market_estimate = yield_accrued + trading_proceeds - curve_fees_paid + flat_fees_returned

yield_accrued      = dy * (c-c0)/c0
trading_proceeds   = dy * (1 - p) * t
curve_fees_paid    = trading_proceeds * curve_fee
flat_fees_returned = dy * t * flat_fee

dy = bond amount
c  = vaultSharePrice (current if non-matured, or checkpoint's if matured)
c0 = openVaultSharePrice
p  = spotPrice
t  = termRemaining (positionDuration - (currentTimestamp - openTimestamp))
```

# Review Checklists

- [ ] **Testing**


MazyGio:
I'm not sure about how we'd want to write tests for these functions, as they're basically estimates. Any direction is appreciated!

crates/hyperdrive-math/src/long/close.rs
```diff
@@ -53,10 +53,54 @@ impl State {
...
+        // get the time remaining
+        let time_remaining = self.calculate_normalized_time_remaining(maturity_time, current_time);
+
+        let flat_value = bond_amount * (fixed!(1e18) - time_remaining);
+        let curve_bonds = bond_amount * time_remaining;
+        let curve_value = curve_bonds * spot_price;
+
+        let trading_proceeds = flat_value + curve_value;
+        let flat_fees_paid = flat_value * self.config.fees.flat.into();
+        let curve_fees_paid = (curve_bonds - curve_value) * self.config.fees.curve.into();
```
dpaiton, line 94: To avoid underflow I recommend checking that `curve_value >= curve_bonds` before this calculation, if so `return Err(eyre!(...))`

crates/hyperdrive-math/src/long/close.rs
```diff
@@ -53,10 +53,54 @@ impl State {
...
+
+        Ok(trading_proceeds - flat_fees_paid - curve_fees_paid)
```
dpaiton, line 96: Same as above, verify that these numbers will work out. You could compute `fees_paid = flat_fees_paid + curve_fees_paid` and then check `trading_proceeds >= fees_paid`. Or do two checks, one for each subtraction.

dpaiton:
I like Mihai's second suggestion, `calculate_market_value_*` better than `calculate_value_*`. But that might just be because I'm not native in this field and thus have a preference for more verbose names. Your call in the end.

crates/hyperdrive-math/src/long/close.rs
```diff
@@ -115,4 +159,52 @@ mod tests {
...
+            let maturity_time = state.position_duration();
+            let current_time = rng.gen_range(fixed!(0)..=maturity_time);
+
+            // Ensure curve_fee is smaller than spot_price to avoid overflows
+            // on the yield space valuation
+            if state.curve_fee() > state.calculate_spot_price()? {
+                continue;
+            }
+
+            let yield_space_valuation = panic::catch_unwind(|| {
```
dpaiton, line 183: there is not a benefit in catching panics and then immediately calling `unwrap`.

The function itself returns `Result<FixedPoint>`, and `panic::catch_unwind` catches any panics from inside the function and bundles them up into another Result, so that the new output type is `Result<Result<FixedPoitn>>`. This is commonly used to allow us to call the function and then handle a broken state gracefully. 

`unwrap` is meant to turn a `Result` into a `panic`. In other words, the outer unwrap is saying if there was a panic in `calculate_close_long` then panic now. The inner unwrap says if there was a `Result` in `calculate_close_long` then panic now. 

I think the idea here is the state is random so we might not be able to close a bond. If so, we want to skip this test. Assuming I'm right, here's what I would suggest we use a match statement to selectively skip this iteration of the test:

```rs
let yield_space_valuation = match panic::catch_unwind(|| { state.calculate_close_long(...) }) {
    Ok(valuation) => match valuation {
        Ok(valuation) => valuation, // no problems calling the fn
        Err(_) => continue, // fn threw an error
     },
    Err(_) => continue, // fn threw a panic
}
```

crates/hyperdrive-math/src/long/close.rs
```diff
@@ -115,4 +159,52 @@ mod tests {
...
+    // Tests market valuation against yield space valuation when closing a long
+    // with the minimum transaction amount.
+    #[tokio::test]
+    async fn test_calculate_value_long() -> Result<()> {
+        let tolerance = int256!(1e15);
+
+        // Fuzz the spot valuation and yield space valuation against each other.
+        let mut rng = thread_rng();
+        for _ in 0..*FAST_FUZZ_RUNS {
+            let state = rng.gen::<State>();
```
dpaiton, line 172: We might want to add a big liquidity amount to this to allow us to reduce the tolerance, since we expect it to be more accurate with higher liquidity. Thankfully, we have a function to do this already!

```rs
            let state = rng.gen::<State>().calculate_pool_state_after_add_liquidity(fixed!(1e27), true);
```

crates/hyperdrive-math/src/long/close.rs
```diff
@@ -115,4 +159,52 @@ mod tests {
...
+
+            let yield_space_valuation = panic::catch_unwind(|| {
+                state.calculate_close_long(bond_amount, maturity_time.into(), current_time.into())
+            })
+            .unwrap()
+            .unwrap();
+
+            let spot_valuation = state
+                .calculate_value_long(bond_amount, maturity_time.into(), current_time.into())
+                .unwrap()
```
dpaiton, line 191: You can use the `?` syntax here instead of `unwrap()`, since the test itself returns a `Result<()>`. I don't _think_ there would actually be a difference here, fwiw, it's just for consistency.

crates/hyperdrive-math/src/long/close.rs
```diff
@@ -53,10 +53,54 @@ impl State {
 
         Ok(flat + curve)
     }
+
+    /// Calculates the market value in shares of a long position using the equation:
+    /// market_estimate = trading_proceeds - flat_fees_paid - curve_fees_paid
+    ///
+    /// trading_proceeds = flat_value + curve_value
+    ///                  = dy * (1 - t) + dy * t * p
```
dpaiton, line 61: You can surround in-line math with a back-tick followed by a dollar-sign ( `` `$` dy * (1-t) `$` `` ) to have it rendered with latex, like elsewhere in the [docs](https://docs.rs/hyperdrive-math/latest/hyperdrive_math/struct.State.html#method.open_long_curve_fee)

crates/hyperdrive-math/src/long/close.rs
```diff
@@ -53,10 +53,54 @@ impl State {
 
         Ok(flat + curve)
     }
+
+    /// Calculates the market value in shares of a long position using the equation:
+    /// market_estimate = trading_proceeds - flat_fees_paid - curve_fees_paid
+    ///
+    /// trading_proceeds = flat_value + curve_value
```
dpaiton, line 60: We have been preferentially rendering math with latex so it is more readable in the [docs page](https://docs.rs/hyperdrive-math/latest/hyperdrive_math/struct.State.html#method.open_long_curve_fee).

dpaiton: Here's an example
```rs
    /// Calculates the market value in shares of a long position.
    ///
    /// market_estimate = trading_proceeds - flat_fees_paid - curve_fees_paid
    /// trading_proceeds = flat_value + curve_value
    /// ```math
    /// \begin{aligned}
    /// \text{trading_proceeds} &= \Delta y \cdot (1 - t) + \Delta y \cdot t \cdot p \\
    ///                  &= \Delta y \cdot ((1 - t) + (t \cdot p)) \\
    /// \text{flat_value} &= \Delta y \cdot (1 - t) \\
    /// \text{curve_bonds} &= \Delta y \cdot t \\
    /// \text{curve_value} &= \text{curve_bonds} \cdot p \\
    /// &= \Delta y \cdot t \cdot p \\
    /// \text{flat_fees_paid} &= \text{flat_value} \cdot \phi_f \\
    /// &= \Delta y \cdot (1 - t) \cdot flat_fee \\
    /// \text{curve_fees_paid} &= (\text{curve_bonds} - \text{curve_value}) \cdot \phi_c \\
    /// &= \Delta y \cdot t \cdot (1 - p) \cdot \phi_c \\
    /// \end{aligned}
    /// ```
    ///
    /// `$\Delta y = \text{bond_amount}$`
    /// `$p  = \text{spot_price}$`
    /// `$t  = \text{time_remaining}$`
    ```

Comment (dpaiton) [reply crates/hyperdrive-math/src/long/close.rs:60]: this is definitely harder to read in-code, which I think is worth considering going forward. We basically have to choose between easier-to-read in our online docs vs easier-to-read in our code. However, we should stick to this for now because that is what we do elsewhere and make the decision to change it later.

**Review by dpaiton — CHANGES_REQUESTED** [review]
Thanks a lot for this, it'll be helpful for several different applications! I made comments on the `long/close.rs` file; prop those to short & I'll give it another look.

File: crates/hyperdrive-math/src/long/close.rs (hunk 7 above)
Comment (MazyGio) [reply crates/hyperdrive-math/src/long/close.rs:60]: Beautiful! Updated.

File: crates/hyperdrive-math/src/long/close.rs (hunk 1 above)
Comment (MazyGio) [reply crates/hyperdrive-math/src/long/close.rs:102]: Thanks for this! I've now added a check for `spot_price <= 1` earlier instead, as that would be the root cause for `curve_value > curve_bonds` to happen.

File: crates/hyperdrive-math/src/long/close.rs (hunk 2 above)
Comment (MazyGio) [reply crates/hyperdrive-math/src/long/close.rs]: Updated.

File: crates/hyperdrive-math/src/long/close.rs (hunk 4 above)
Comment (MazyGio) [reply crates/hyperdrive-math/src/long/close.rs]: Ohhh nice! Tried this out, decreased tolerance to 1e13. The error is taken as a percentage, so that's a 0.001% difference between spot and yield space valuations.

File: crates/hyperdrive-math/src/long/close.rs (hunk 3 above)
Comment (MazyGio) [reply crates/hyperdrive-math/src/long/close.rs]: I looked into this again (with a much better understanding thanks to your comments), and realized that most of what I had there wasn't necessary. We're trying to close the `minimum_transaction_amount` of bonds and, while the state is random, it's always a valid state, with enough breathing room that we would be able to close those bonds. 

There are some possible states that cause a `panic` in `calculate_close_long`, but lines 179-181 specifically skip those. Is that alright or would it be better practice to more generically catch panics and skip them here instead?

File: crates/hyperdrive-math/src/long/close.rs (hunk 5 above)
Comment (MazyGio) [reply crates/hyperdrive-math/src/long/close.rs]: Fantastic! Updated.

File: crates/hyperdrive-math/src/long/close.rs [hunk 8]
```diff
@@ -115,4 +172,56 @@ mod tests {
...
+            let current_time = rng.gen_range(fixed!(0)..=maturity_time);
+
+            // Ensure curve_fee is smaller than spot_price to avoid overflows
+            // on the yield space valuation, as that'd mean having to pay a larger
+            // amount of fees than the current value of the long
+            if state.curve_fee() > state.calculate_spot_price()? {
+                continue;
+            }
+
+            let yield_space_valuation = state.calculate_close_long(
```
dpaiton, line 199: I think calling this `yield_space_valuation` is misleading, because it includes fees and flat + curve. It's really the hyperdrive valuation. Same in the short test.

dpaiton (approved):
Awesome, thanks for building this! Left a minor comment to change a variable name, otherwise it looks great.

MazyGio:
`test_calculate_market_value_short` passes with a tolerance of 1e12. It's worth noting that the test attempts to close a short of size `minimum_transaction_amount` (range from 0.1 to 1 bonds).

There's a couple things I've been able to drill down on while testing this function:
1. I've reworked the function to match `calculate_close_long` as closely as possible. The only difference is now on how both functions calculate the curve part of the trade. This makes up the entire error amount on this test.
2. Using a relatively large liquidity amount (1e27) by doing `.calculate_pool_state_after_add_liquidity(fixed!(1e27), true)?;` after the initial State is generated does result in a smaller `market_impact`, but it somehow creates a larger discrepancy between `hyperdrive_valuation` and `spot_valuation`. I can't say for certain why that is, but it makes the `hyperdrive_valuation` result a bit erratic. I'd expect it to be slightly lower than the `spot_valuation`, since the `hyperdrive_valuation` considers market impact, but sometimes it's not.

I suspect small rounding-related inaccuracies when computing `calculate_shares_in_given_bonds_out_up_safe`, made relatively large due to the large share reserve and adjustment levels compared to the small position size.

Example failure using `.calculate_pool_state_after_add_liquidity(fixed!(1e27), true)?;`:
```
==========================================
spot_price:       0.073223040957386746
spot_price_after: 0.073223040957595844
market_impact:        209098         <--[O(1e7). very small!]
valuation_error:      11363446861    <--[O(1e11). pretty large]
spot_valuation:       0.111224043778463171
hyperdrive_valuation: 0.111224055141910032    <--[larger than spot, which is weird]
share_reserves:            676920352.732639701955744293
share_adjustment:          -15427047685553773566980402
effective_share_reserves:  692347400.418193475522724695
failed 603 times.    <--[out of 10k]
==========================================
test short::close::tests::test_calculate_market_value_short ... ok
```

Example failure without that line:
```
==========================================
spot_price:       0.353913457944902287
spot_price_after: 0.353913875316959714
market_impact:        417372057427   <--[O(1e12)]
valuation_error:      54535898369    <--[O(1e11). Same order of magnitude as above]
spot_valuation:       0.726982046773411663
hyperdrive_valuation: 0.726981992237513294    <--[when it's higher than spot, the error is smaller than 1e10]
share_reserves:            95894096.977191008442078684
share_adjustment:          95886552995656698674111646
effective_share_reserves:  7543.981534309767967038
failed 17 times.    <--[out of 10k]
==========================================
```

Both tests used a tolerance of 1e10 (to ensure it failed a bunch of times)

3. When I don't use the larger liquidity amount, the error is always within one order of magnitude from the price impact. It's worth noting that, in lieu of a `calculate_spot_price_after_close_short` function, I've approximated the price impact by using `calculate_spot_price_after_long`, using the amount of non-matured bonds, which should have a fairly similar effect.

Would we be okay with the tests passing with "large" discrepancies between `hyperdrive_valuation` and `spot_valuation`, as long as these discrepancies are e.g. within one or two orders of magnitude of the position's market impact?

MazyGio:
After reworking these to match their `calculate_close_` counterparts, I've tested extensively to figure out a base tolerance, which we can scale depending on the ratio of `effective_share_reserves` relative to `bond_reserves`. When this ratio `ze / y` is too low, it creates a large discrepancy due to the `market_impact` that affects the hyperdrive valuation, so we bump the base tolerance of `1e12` to `1e13` in such cases.

Thanks @dpaiton for all the help in getting these completed!

crates/hyperdrive-math/src/short/close.rs
```diff
@@ -418,4 +477,68 @@ mod tests {
...
+            let bond_amount = state.minimum_transaction_amount();
+            let open_vault_share_price = rng.gen_range(fixed!(0.5e18)..=fixed!(2.5e18));
+            let maturity_time = U256::try_from(state.position_duration())?;
+            let current_time = rng.gen_range(fixed!(0)..=FixedPoint::from(maturity_time));
+
+            // When the reserves ratio is too small, the market impact makes the error between
+            // the valuations larger, so we scale the test's tolerance up to make up for it,
+            // since this is meant to be an estimate that ignores liquidity constraints.
+            let reserves_ratio = state.effective_share_reserves()? / state.bond_reserves();
+            if reserves_ratio < fixed!(1e12) {
```
dpaiton, line 505: I think you want to nest these, unless the intention is to have the scale be 1,000 if it is <1e14? If so, I'd be more explicit. So either

```
if reserves_ratio >= fixed!(1e14) && reserves_ratio < fixed!(1e12) {
    scaled_tolerance *= int256!(100); // 1e14
}
else if reserves_ratio < fixed!(1e14) {
    scaled_tolerance *= int256!(10); // 1e13
}
```

or

```
if reserves_ratio >= fixed!(1e14) && reserves_ratio < fixed!(1e12) {
    scaled_tolerance *= int256!(100); // 1e14
}
else if reserves_ratio < fixed!(1e14) {
    scaled_tolerance *= int256!(1_000); // 1e15
}
```

MazyGio: You're absolutely right! They were meant to be nested. Excellent catch, pushing it now, merging once CI passes!

PR #160, opened 2024-07-03
update cargo publish path

dpaiton:
fixes publish to use a different path spec for the crates.


scripts/publish_crates.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
 echo "publish rust crates to crates.io"
-cargo publish --token $1 -p file:://${PWD}/crates/fixedpointmath
-cargo publish --token $1 -p file:://${PWD}/crates/hyperdrive-wrappers
-cargo publish --token $1 -p file:://${PWD}/crates/hyperdrive-math
\ No newline at end of file
+cargo publish --token $1 -p fixedpointmath
+BUILD_DISABLED=true cargo publish --token $1 -p hyperdrive-wrappers
```
ryangoree, line 5:
```suggestion
cargo publish --token $1 -p hyperdrive-wrappers
```

I think we can delete this now since the build script isn't in the manifest anymore.

ryangoree (approved):

scripts/publish_crates.sh
dpaiton: I'm not sure why, but I needed this flag locally anyway. I think it's best to leave it in JIC; it's not hurting anything.

ryangoree (approved):

PR #161, opened 2024-07-03
fix publish doc build

dpaiton:
Make a copy of the html header into a new assets folder this time.

### DIFF
--- a/scripts/publish_crates.sh
+++ b/scripts/publish_crates.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+echo "simlink assets into crates"
+ln -s assets crates/fixedpointmath
+ln -s assets crates/hyperdrive-wrappers
+ln -s assets crates/hyperdrive-math
+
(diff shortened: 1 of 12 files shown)


scripts/publish_crates.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+echo "simlink assets into crates"
+ln -s assets crates/fixedpointmath
+ln -s assets crates/hyperdrive-wrappers
+ln -s assets crates/hyperdrive-math
+
```
ryangoree, line 7: I think we'll need to copy here.

ryangoree (approved):

scripts/publish_crates.sh
```diff
@@ -1,6 +1,16 @@
 #!/bin/bash
 
+echo "simlink assets into crates"
```
ryangoree, line 3:
```suggestion
echo "copy assets into crates"
```

ryangoree (approved):

PR #168, opened 2024-07-09
Change `calculate_solvency` to return a `Result<FixedPoint>`

dpaiton:
# Description
- changed the `calculate_solvency` function to return a `Result<FixedPoint>` to throw a catchable error if the pool is not solvent.


MazyGio (approved):
Looks great!

ryangoree (approved):

PR #170, opened 2024-07-10
Inclusive & valid random share reserve range; dynamic fuzz bugfix

dpaiton:
# Description
- share adjustment (zeta) was using a non-inclusive range, but the requirement (`<=`) implies inclusive should be valid.
  - I tested this with the bounds (`share_adjustment = 0` and `share_adjustment = share_reserves - min_share_reserves - fixed!(10e18)`) and all tests passed with both extremes
  - note that I had to reduce the upper-bound by `fixed!(10e18)` for the tests to pass, which I guess means we were just getting lucky before. I made an issue for this: https://github.com/delvtech/hyperdrive-rs/issues/171
- dynamic fuzz is supposed to do 10x fuzz runs on new tests, but was 1x with `FAST_FUZZ_RUNS`
- fix a check in an `open.rs` test where the allowable max bond amount was too low

### DIFF
--- a/.github/workflows/dynamic_fuzz.yml
+++ b/.github/workflows/dynamic_fuzz.yml
@@ -57,7 +57,7 @@ jobs:
             echo "$new_tests"
             while IFS= read -r test; do
               echo "Running test: $test"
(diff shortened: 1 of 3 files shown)


ryangoree (approved):
🚀 (pending missing `=` commit)

Issue #171, opened 2024-07-11
Tests fail when using large (but valid) `share_adjustment`


dpaiton:
https://github.com/delvtech/hyperdrive-rs/pull/170 highlights that using the upper-bound on `share_adjustment` when generating a random state causes a set of tests to fail:

```rs
test lp::math::tests::fuzz_calculate_net_curve_trade_safe
test lp::math::tests::fuzz_calculate_present_value
test long::max::tests::fuzz_sol_calculate_max_long
test short::open::tests::test_defaults_calculate_spot_price_after_short
```

After an initial inspection, @jalextowle concluded that at least the `fuzz_sol_calculate_max_long`, `fuzz_calculate_present_value`, and `fuzz_calculate_net_curve_trade_safe` failures indicate a discrepancy between Rust and Solidity.

We need to dig into each test and correct it or the underlying code it is testing.

PR #178, opened 2024-07-16
fix open long delta calculations and add tests

dpaiton:
# Resolved Issues
working towards https://github.com/delvtech/hyperdrive-rs/issues/171 and https://github.com/delvtech/hyperdrive-rs/issues/21

# Description
- The code for calculating pool deltas & state after opening a long did not correctly account for the gov fee's impact on the share reserves. I fixed this and added a test to demonstrate the failure
- I added a similar test on the short side, which almost passes with equality (seems there is a rounding error)
- I renamed some functions, cleaned up some comments, fixed some typos


ryangoree (approved):
reviewed in #180

PR #184, opened 2024-07-29
Wip FixedPoint refactor

### LINKED ISSUE #127: Change FixedPoint to only support From<U128> and adjust tests
Modify the [U256 from](https://github.com/delvtech/hyperdrive-rs/blob/83b2f812f859b23b3bdc7401b9c421e8786c2c46/crates/fixed-point/src/lib.rs#L67-L72) impl to check if the value fits in a U128. If so, do the conversion. Else (value is > U128::max) throw an error.

Adjust the tests to use U128::max instead of U256::max everywhere. 

We should make sure to add docstrings somewhere to indicate to the user that FixedPoint has max U128 support.

ryangoree:
# Resolved Issues

closes #127 

# Description


crates/fixedpointmath/src/value.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    fn flip_sign_if(self, condition: bool) -> Self {
+        if condition {
+            return self.flip_sign();
+        }
+        self
+    }
+}
+
+// TODO: On hold until varying decimal precision is implemented
```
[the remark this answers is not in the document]
ryangoree: I wasn't sure yet, so I just commented it out and left places to implement it later if we feel it's needed.

crates/fixedpointmath/src/lib.rs
```diff
@@ -1,422 +1,406 @@
...
 pub use utils::*;
+pub use value::*;
 
-/// A fixed point wrapper around the `U256` type from ethers-rs.
+/// A fixed point wrapper around ethers-rs.
+///
+/// This fixed point type is a heavily based on Solidity's FixedPointMath
+/// library with a few changes:
+/// - Support for negative numbers.
+/// - Support for overflowing intermediate operations in `mul_div_down` and
```
jalextowle, line 31: Is this on by default? This seems like an anti-pattern for use-cases where we are trying to achieve parity

crates/fixedpointmath/src/lib.rs
```diff
@@ -1,422 +1,406 @@
...
-    fn default() -> FixedPoint {
-        fixed!(0)
-    }
+/// ensure that the behavior is identical given values bounded by both the
+/// Solidity and Rust limits.
+#[derive(PartialOrd, Ord, Clone, Copy)]
+pub struct FixedPoint<T: FixedPointValue> {
+    sign: FixedPointSign,
+    value: T,
+    decimals: u8,
```
jalextowle, line 41: Why are we storing this with the value? It’s just copying around data that won’t be used unless we display it.

jalextowle: Nvm, I see that it’s being used to support fixed point addition with a different decimal point. We could do this before mulDivDown (since it’s just a*b/c), but I see how this could be nicer in some cases

crates/fixedpointmath/src/value.rs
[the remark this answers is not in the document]
jalextowle: Mind if we just make it an enhancement issue? That way the commented code doesn’t have to be checked in

crates/fixedpointmath/src/lib.rs
ryangoree: Yes, we already reached parity, but now I'm going past it to address pitfalls in the solidity lib. Then we'll switch to using `FixedPoint<u128>` and `FixedPoint<i128>` in all the tests so that we can be sure the inputs are bounded by those integer types which we know are within ranges the Solidity can support. This way we stop running into errors related to bad inputs and focus on protocol correctness.

crates/fixedpointmath/src/value.rs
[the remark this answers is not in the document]
ryangoree: Yeah, I have a tendency to keep my notes in my code so it's always in my face, but it's not a good practice in shared code bases 😓

crates/fixedpointmath/src/lib.rs
jalextowle: Is the idea that the inputs are u128s but that the intermediate multiplications are done as U256? That makes sense to me. 
If so, this isn’t really a problem with the Solidity library. It’s more an issue with the sampling that we’re currently doing.

ryangoree: Yeah I realized that as I was half way into the refactor. That this is really just a sampling issue, but I was already deep into the refactor and this will make the lib more flexible beyond parity with the lib used in Hyperdrive.

bindings/hyperdrivepy/src/hyperdrive_state_methods/long/open.rs
```diff
@@ -42,8 +49,20 @@ impl HyperdriveState {
                     err
                 ))
             })?;
-        let share_result = U256::from(share_result_fp).to_string();
-        let bond_result = U256::from(bond_result_fp).to_string();
+        let share_result = U256::try_from(share_result_fp)
+        .map_err(|err| {
+            PyErr::new::<PyValueError, _>(format!(
+                "Failed to convert calculate_pool_deltas_after_open_long.0 (share_reserves) result {} to U256: {}",
```
dpaiton, line 55: what is the "0" indicating?

bindings/hyperdrivepy/src/hyperdrive_state_methods/lp/add.rs
```diff
@@ -98,9 +105,23 @@ impl HyperdriveState {
                     err
                 ))
             })?;
-        let result1 = U256::from(result_fp1).to_string();
+        let result1 = U256::try_from(result_fp1)
+            .map_err(|err| {
+                PyErr::new::<PyValueError, _>(format!(
+                "Failed to convert calculate_pool_deltas_after_add_liquidity.0 (share_reserves) result {} to U256: {}",
```
dpaiton, line 111: again, since this is going to be seen in a python environment I think `[0]` notation makes more sense.

bindings/hyperdrivepy/src/hyperdrive_state_methods/lp/remove.rs
```diff
@@ -102,8 +102,21 @@ impl HyperdriveState {
         let (withdrawal_shares_redeemed, share_proceeds, _state) = result;
 
         Ok((
-            U256::from(share_proceeds).to_string(),
-            U256::from(withdrawal_shares_redeemed).to_string(),
+            U256::try_from(share_proceeds)
+                .map_err(|err| {
+                    PyErr::new::<PyValueError, _>(format!(
+            "Failed to convert calculate_remove_liquidity.share_proceeds result {} to U256: {}",
```
dpaiton, line 108: I actually like this syntax better than the index used above; what do you think about using it across the board?

crates/fixedpointmath/src/lib.rs
```diff
@@ -1,422 +1,406 @@
...
+        S: Into<FixedPointSign>,
+        V: Into<T>,
+    {
+        let sign = sign.into();
+        let value = value.into();
+
+        if (sign == FixedPointSign::Positive && value > T::MAX)
+            || (sign == FixedPointSign::Negative && value < T::MIN)
+        {
+            bail!(
```
dpaiton, line 84: bail is cool! I didn't know about that

crates/fixedpointmath/src/lib.rs
```diff
@@ -1,422 +1,406 @@
...
-        let mut ylnx: I256 = y_int256.wrapping_mul(lnx);
-        ylnx = ylnx.wrapping_div(int256!(1e18));
+        // If the exponent is negative, return 1 / x^abs(y).
+        if y.is_negative() {
+            let abs_result = self.pow(y.abs())?;
 
-        // Calculate exp(y * ln(x)) to get x^y
-        Self::exp(ylnx)?.try_into()
-    }
+            // FIXME: What's the correct return value here?
```
dpaiton, line 355: python float handles this by including special values `inf` and `-inf` and `nan`. https://github.com/delvtech/fixedpointmath/blob/main/fixedpointmath/fixed_point.py#L164
then we check that member variable whenever people want to perform an operation.
You can see all of the rules here: https://github.com/delvtech/fixedpointmath/blob/main/tests/test_fp_nonfinite_checks.py

crates/hyperdrive-math/src/long/max.rs
```diff
@@ -574,7 +574,17 @@ mod tests {
                         FixedPoint::from(expected_base_amount)
                     );
                 }
-                Err(_) => assert!(actual.is_err() || actual.unwrap().is_err()),
+                Err(e) => {
+                    if !actual.is_err() {
+                        let unwrapped = actual.unwrap();
```
dpaiton, line 579: is this right? My understanding is that unwrap does implicit error handling and will either return the inner element or panic. So you'll never get to the assert if it is an error.

crates/hyperdrive-math/src/base.rs
```diff
@@ -57,15 +55,15 @@ impl State {
         &self,
         scaled_maturity_time: FixedPoint,
         current_time: U256,
-    ) -> FixedPoint {
+    ) -> Result<FixedPoint> {
         let scaled_latest_checkpoint =
-            FixedPoint::from(self.to_checkpoint(current_time)) * fixed!(1e36);
-        let scaled_position_duration = self.position_duration() * fixed!(1e36);
+            FixedPoint::from(self.to_checkpoint(current_time)).mul_down(fixed!(1e36))?;
```
dpaiton, line 60: did we lose the `*` sugar or are you preferring `mul_down`? We've been somewhat inconsistent with this, but imo the sugar operators (e.g. `*`) are better unless we're explicitly trying to match a solidity function because they are more readable.

crates/hyperdrive-math/src/base.rs
```diff
@@ -1,18 +1,15 @@
...
 impl State {
     /// Calculates the number of share reserves that are not reserved by open
     /// positions.
-    pub fn calculate_idle_share_reserves(&self) -> FixedPoint {
-        let long_exposure = self.long_exposure().div_up(self.vault_share_price());
-        match self.share_reserves() > long_exposure + self.minimum_share_reserves() {
-            true => self.share_reserves() - long_exposure - self.minimum_share_reserves(),
-            false => fixed!(0),
-        }
+    // FIXME: This function is redundant with `calculate_solvency`.
```
dpaiton, line 10: I agree that we should only have one function, but I think `calculate_idle_share_reserves` is more clear than `calculate_solvency`. The former is more explicit in what is being computed. 

If we allowed negative idle share reserves, then we could also implement a `check_solvency` function that returns a bool based on the sign of `calculate_idle_share_reserves`. Or we do not allow negative idle share reserves and leave it to the developer to know that this is the function to run to check & calculate solvency.

crates/fixedpointmath/src/math.rs
```diff
@@ -0,0 +1,459 @@
...
+                    }
+                }
+            }
+        )*
+    };
+    ($($tt:tt)*) => {};
+}
+
+// Forward these operators to the underlying `FixedPointValue`.
+impl_forwarded_operator!(Add, Sub, Rem);
```
ryangoree, line 205: As a consequence of forwarding these operators, they will only work with 2 FixedPoint instances that have the same underlying value type. In the future, we can implement functions to make them more flexible like the rest of the ops.

crates/fixedpointmath/src/lib.rs
```diff
@@ -1,859 +1,35 @@
+//! A generic fixed point wrapper around the `U256` type from ethers-rs.
```
ryangoree, line 1: Since `ethers-rs` is being deprecated, I checked out alloy to see what they use for ints now and they're using [ruint](https://crates.io/crates/ruint) which looks nice. Also found [uint](https://crates.io/crates/uint). It might be nice to refactor to something similar soon.

crates/hyperdrive-math/src/base.rs
ryangoree: Reverted for now 👍

ryangoree:
closed in favor of #186

ryangoree:
closed in favor of #186

PR #186, opened 2024-08-12
Refactor `fixedpointmath`

ryangoree:
Related to: #127

(replaces wip pr: #184)

# Description

This refactors the `FixedPoint` type to be generic with implementations for `i128`, `u128`, `I256`, and `U256`. For the most part, the library can be used the same, but with a type param, eg:

```diff
- fn foo() -> FixedPoint {}
+ fn foo() -> FixedPoint<U256> {}
```

However, the API has been expanded and there's plenty code reduction we can do in future PRs. It's probably also worth rethinking argument types now that the library can support negative numbers and be bound by different limits.

## Expanded API

I wrote doc comments throughout to aid in adoption, but below is a skeleton containing most of the new API to get a quick sense:

```rs
pub struct FixedPoint<T: FixedPointValue> {
    raw: T,
    decimals: u8,
}

impl<T: FixedPointValue> FixedPoint<T> {
    pub const MIN: Self;
    pub const MAX: Self;

    // Constructors //
    pub fn new<V: Into<T>>(value: V) -> Self;
    pub fn try_from<V: TryInto<T> + Debug>(value: V) -> Result<Self>;
    pub fn from_sign_and_abs(sign: FixedPointSign, abs: U256) -> Result<Self>;
    pub fn from_dec_str(s: &str) -> Result<Self>;
    pub fn saturate_sign(sign: FixedPointSign) -> Self;
    pub fn zero() -> Self;
    pub fn one(&self) -> Self;

    // Getters //
    pub fn raw(&self) -> T;
    pub fn decimals(&self) -> u8;
    pub fn sign(&self) -> FixedPointSign;

    // Predicates //
    pub fn is_negative(&self) -> bool;
    pub fn is_positive(&self) -> bool;
    pub fn is_zero(&self) -> bool;

    // Conversion to other FixedPoint types //
    pub fn change_type<U: FixedPointValue + TryFrom<T>>(self) -> Result<FixedPoint<U>>;

    // Conversion to unsigned & signed ethers types //
    pub fn to_u256(self) -> Result<U256>;
    pub fn to_i256(self) -> Result<I256>;

    // Conversion to unsigned and signed std types //
    pub fn to_u128(self) -> Result<u128>;
    pub fn to_i128(self) -> Result<i128>;

    // Formatting //
    pub fn to_scaled_string(&self) -> String;

    // Math //
    pub fn abs(&self) -> Self;
    pub fn unsigned_abs(&self) -> FixedPoint<U256>; // `U256` to avoid overflow.
    pub fn abs_diff(&self, other: Self) -> FixedPoint<U256>;
    pub fn mul_div_down(self, other: Self, divisor: Self) -> Self;
    pub fn mul_div_up(self, other: Self, divisor: Self) -> Self;
    pub fn mul_down(self, other: Self) -> Self;
    pub fn mul_up(self, other: Self) -> Self;
    pub fn div_down(self, other: Self) -> Self;
    pub fn div_up(self, other: Self) -> Self;
    pub fn pow(self, y: Self) -> Result<Self>;
}

// Direct conversions between primitive types and FixedPoint for any
// `FixedPointValue` that can be converted to and from the primitive type.
conversion_impls!(i8, u8, i16, u16, i32, u32, i64, u64, isize, usize, [u8; 32], bool);

// Value //

/// A value that can be used to perform fixed-point math.
pub trait FixedPointValue: /* ... */ {
    const MIN: Self;
    const MAX: Self;
    const MAX_DECIMALS: u8 = 18;

    fn is_signed() -> bool;
    fn is_negative(&self) -> bool;
    fn is_positive(&self) -> bool;
    fn is_zero(&self) -> bool;
    fn flip_sign(self) -> Self;
    fn flip_sign_if(self, condition: bool) -> Self;
    fn abs(self) -> Self;
    fn unsigned_abs(self) -> U256;

    pub fn from_u256(u: U256) -> Result<Self>;
    pub fn from_u128(u: u128) -> Result<Self>;
    
    pub fn to_u256(self) -> Result<U256>;
    pub fn to_u128(self) -> Result<u128>;
}

// Conversions traits //

pub trait Fixed: FixedPointValue {
    fn fixed(self) -> FixedPoint<Self>;
}

// Add `.fixed()` to all types that implement `FixedPointValue`.
impl<T: FixedPointValue> Fixed for T {}

pub trait ToFixed: Sized + Debug {
    fn to_fixed<T: FixedPointValue + From<Self>>(self) -> FixedPoint<T>;
    fn try_to_fixed<T: FixedPointValue + TryFrom<Self>>(self) -> Result<FixedPoint<T>>;
}

// Add `.to_fixed()` & `.try_to_fixed()` to all sized types that implement
// `Debug`.
impl<T: Sized + Debug> ToFixed for T {}

// Macros //

macro_rules! uint256 {}    // -> `U256`
macro_rules! int256 {}     // -> `I256`
macro_rules! fixed {}      // -> `FixedPoint<T>` (where `T` is inferred)
macro_rules! fixed_u256 {} // -> `FixedPoint<U256>`
macro_rules! fixed_i256 {} // -> `FixedPoint<I256>`
macro_rules! fixed_u128 {} // -> `FixedPoint<u128>`
macro_rules! fixed_i128 {} // -> `FixedPoint<i128>`
```


dpaiton (changes requested):
Partial review; will pick it back up shortly.

crates/fixedpointmath/src/fixed_point.rs
```diff
@@ -0,0 +1,399 @@
...
+        Ok(Self::new(value))
+    }
+
+    pub fn from_sign_and_abs(sign: FixedPointSign, abs: U256) -> Result<Self> {
+        Ok(match sign {
+            FixedPointSign::Positive => Self::new(T::from_u256(abs)?),
+            FixedPointSign::Negative => {
+                if abs == T::MIN.unsigned_abs() {
+                    // Avoid overflow
+                    Self::MIN
```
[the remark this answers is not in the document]
ryangoree: Trying to call `T::from_u256(T::MIN.unsigned_abs())` would overflow since the
absolute `MIN` value of a two's compliment integer is 1 more than its `MAX`, so instead
of creating the positive value first, then flipping the sign, I'm just
shortcutting to the actual value.

I'll add a comment for this: 👍

crates/fixedpointmath/src/fixed_point.rs
```diff
@@ -0,0 +1,399 @@
...
+                } else {
+                    let raw = T::from_u256(abs)?.flip_sign();
+                    Self::new(raw)
+                }
+            }
+        })
+    }
+
+    pub fn from_dec_str(s: &str) -> Result<Self> {
```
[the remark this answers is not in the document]
ryangoree: Ah, yeah. I had originally planned to support adding signs to unsigned integers (e.g., a negative `U256`), but:

- Keeping the raw value sign in sync with the outer type gets messy and inneficient.
- A negative `U256` that's too low to be converted to a `I256` is kinda hard to work with since the `raw` value can't be represented using the underlying type.
- Lastly, two's compliment is already a great and efficient system for representing signed integers; no need for me to get *too* clever.

Instead, the lib now inherits the sign from the underlying type. If you plan to use negative numbers, you'd use an `i128` or `I256`. This way the behavior is predictable and calling `raw()` will return the actual underlying value, sign and all.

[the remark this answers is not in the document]
ryangoree: Ah, yeah. I had originally planned to support adding signs to unsigned integers (e.g., a negative `U256`), but:

- Keeping the raw value sign in sync with the outer type gets messy and inneficient.
- A negative `U256` that's too low to be converted to a `I256` is kinda hard to work with since the `raw` value can't be represented using the underlying type.
- Lastly, two's compliment is already a great and efficient system for representing signed integers; no need for me to get *too* clever.

Instead, the lib now inherits the sign from the underlying type. If you plan to use negative numbers, you'd use an `i128` or `I256`. This way the behavior is predictable and calling `raw()` will return the actual underlying value, sign and all.

crates/hyperdrive-math/src/long/targeted.rs
```diff
@@ -215,12 +219,12 @@ impl State {
     /// r'(x) = \frac{-p'(x)}{t \cdot p(x)^2}
     /// ```
     ///
-    /// We return `$-r'(x)$` because negative numbers cannot be represented by FixedPoint.
+    /// We return `$-r'(x)$` because negative numbers cannot be represented by FixedPoint<U256>.
```
ryangoree, line 222: I did a find and replace for `FixedPoint` to `FixedPoint<U256>` at the start, so there will be some comments like this that aren't true anymore, but I figure those can be addressed in follow-up PRs along with the code the comment refers to.

crates/fixedpointmath/src/fixed_point.rs
```diff
@@ -0,0 +1,399 @@
+use std::fmt::{self, Debug};
+
+use ethers::types::{I256, U256};
+use eyre::{bail, eyre, Result};
+
+use crate::{sign::FixedPointSign, utils::u256_from_str, value::FixedPointValue};
+
+/// A generic fixed point wrapper around the `U256` type from ethers-rs.
```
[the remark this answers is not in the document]
ryangoree: Not really.

crates/fixedpointmath/src/math.rs
```diff
@@ -0,0 +1,450 @@
...
+    #[tokio::test]
+    async fn fuzz_mul_div_down() -> Result<()> {
+        let chain = Chain::connect(None, None).await?;
+        chain.deal(DEPLOYER.address(), uint256!(100_000e18)).await?;
+        let client = chain.client(DEPLOYER.clone()).await?;
+        let mock_fixed_point_math = MockFixedPointMath::deploy(client, ())?.send().await?;
+
+        // Fuzz the rust and solidity implementations against each other.
+        let mut rng = thread_rng();
+        for _ in 0..10_000 {
```
dpaiton, line 240: can you use the FUZZ constant here & elsewhere?

crates/fixedpointmath/src/math.rs
```diff
@@ -0,0 +1,450 @@
...
+    pub fn pow(self, y: Self) -> Result<Self> {
+        let one = self.one();
+
+        // If the exponent is negative, return 1 / x^abs(y).
+        if y.is_negative() {
+            let abs_result = self.pow(y.abs())?;
+
+            // FIXME: What's the correct return value here?
+            if abs_result.is_zero() {
+                return Ok(Self::zero());
```
dpaiton, line 94: I would return an error here, since we're dividing by zero and that should be `nan`

dpaiton (approved):
Looks good besides the nits! Thanks for this; it looks like it was a big effort but it is definitely going to be worth it :)

crates/fixedpointmath/src/math.rs
ryangoree: Sounds good. I'm going to remove the check for `0` then for now and let the
error from the `div_down` below propagate until we decide a diff error or return
val here.

crates/fixedpointmath/src/math.rs
ryangoree: Oh, I remember why the `const` isn't used here. It comes from the `hyperdrive-test-utils` crate which isn't a dependency of `fixedpointmath` (because it already depends on `fixedpointmath` itself).

I'm going to leave these for now. But maybe we reconsider the env variable names for all constants across crates, so they can use their own const defs (without depending on eachother), but still share the same values by referencing the same env vars.

Issue #185, opened 2024-08-06
Inconsistency when using `calculate_open_short` and `calculate_max_short` together


DannyDelott:
#### Background

The sdk's `calculate_max_short` method takes a `budget` and returns a bond amount. 

The `calculate_open_short` takes a bond amount and returns the trader deposit (ie: `budget`). 

We expect that the outputs of one would reflect the inputs of the other, however the steps below show that this is not the case.

#### Steps to reproduce

1. Call `calculate_max_short` with a max uint256 budget to get the pool's max short (in bonds).
2. Feed this bond amount into `calculate_open_short` to get the actual `traderDeposit` needed.
3. Feed the `traderDeposit` back into `calculate_max_short` as the budget.

#### Expected
The bond amount returned from `calculate_max_short` should be the same as the bond amount we gave to `calculate_open_short`.

#### Actual
Feeding the `traderDeposit` back into `calculate_max_short` results in significantly fewer bonds shorted.

Usecase:
In the Open Short form, users can input either an amount to pay (budget) or the number of bonds to short. 

Depending on which one they type into, either `calculate_max_short` or `calculate_open_short` will be called to update the form.

![image](https://github.com/user-attachments/assets/2ccefd2b-68f5-4e7b-b210-2a8d4068902f)

DannyDelott:
We're currently not passing a conservative_price argument to `calcMaxShort`, there is prior art on how to calculate this here that we should try:

https://github.com/delvtech/hyperdrive-rs/blob/e4fcc3994a8ec3cf657de418309c0fe9909b3b00/crates/hyperdrive-math/src/test_utils/agent.rs#L178

dpaiton:
Tasks :

- [ ] fix `max_short_guess` https://github.com/delvtech/hyperdrive-rs/pull/198
- [ ] fix `calculate_absolute_max_short` 
- [ ] fix `calculate_max_short` https://github.com/delvtech/hyperdrive-rs/pull/199
- [ ] fix or deprecate conservative LP price calculation
- [ ] write wasm & python bindings

PR #199, opened 2024-10-02
add `calculate_short_bonds_given_deposit`

dpaiton:
# Resolved Issues
working towards https://github.com/delvtech/hyperdrive-rs/issues/185
Also makes more progress towards https://github.com/delvtech/hyperdrive-rs/issues/21

# Description
This provides an alternative function to `calculate_max_short` that includes a tolerance parameter and guarantees* convergence with sufficient number of iterations.

I changed the signature & behavior from `calculate_max_short` in these ways:
- Most importantly, we now ask for the absolute max  short (aka pool's max regardless of budget) as an input argument. This is to make clear to the user that there are two required iterative processes to get this answer: first one is to find the absolute max short, second is to find a max short relative to a budget.
- This function should also not be used with an unlimited budget to find the max short. Use `calculate_absolute_max_short` for that. If you provide a budget that is larger than the max possible, then this function will throw an error.
- The result is guaranteed to be less than or equal to your budget; it will not overshoot the target base amount even if the result is within tolerance and solvent.

 *I tested a lot of options, but I don't have any formal guarantees on convergence. The most extreme values I tried was `1e5` tolerance with `1M` steps. These two knobs can be tuned as needed for trading off compute speed against tolerance.

# Remaining tasks for future PRs
- The new test is pretty janky because of the need to put several checks around calculating the absolute max bond amount. That function needs to be investigated next.

- I still need to make the wasm & python bindings. This should be done after I fix up absolute max bond amount.

- I created this as a new function because I want to give it a chance to exist on its own and be tested before we deprecate the existing `calculate_max_short` function. Once we have battle tested it, we will want to do a cleanup.

- The method for calculating the conservative price is still not guaranteed to be safe.

### DIFF
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -88,8 +88,6 @@ impl State {
                 self.calculate_open_short(last_good_bond_amount, open_vault_share_price)?;
 
             // Calculate the current loss.
-            // TODO: Loss should **always** go down, but it does not.
-            // Ignoring because fuzz tests are passing; but need to investigate.
             loss = if current_base_amount < target_base_amount {
                 // It's possible that a nudge from failure cases in the previous
                 // iteration put us within the tolerance.
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -1,6 +1,6 @@
 use ethers::types::{I256, U256};
 use eyre::{eyre, Result};
-use fixedpointmath::{fixed, FixedPoint};
+use fixedpointmath::{fixed, fixed_u256, FixedPoint};
 
 use crate::{calculate_effective_share_reserves, State, YieldSpace};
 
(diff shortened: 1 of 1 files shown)


crates/hyperdrive-math/src/short/max.rs
```diff
@@ -38,6 +38,135 @@ impl State {
...
+
+        // Run Newton's Method to adjust the bond amount.
+        let mut loss = FixedPoint::from(U256::MAX);
+        for _ in 0..max_iterations {
+            // Calculate the current deposit.
+            let current_base_amount =
+                self.calculate_open_short(last_good_bond_amount, open_vault_share_price)?;
+
+            // Calculate the current loss.
+            // TODO: Loss should **always** go down, but it does not.
```
jrhea, line 91: Should this be an issue?

crates/hyperdrive-math/src/short/max.rs
```diff
@@ -568,9 +697,102 @@ mod tests {
...
+            let state = rng.gen::<State>();
+            let checkpoint_exposure = {
+                let value = rng.gen_range(fixed!(0)..=FixedPoint::from(U256::from(U128::MAX)));
+                if rng.gen() {
+                    -I256::try_from(value)?
+                } else {
+                    I256::try_from(value)?
+                }
+            };
+            // TODO: Absolute max doesn't always work. Sometimes the random
```
jrhea, line 719: Does this need to go into an issue, or are you planning to fix this in a follow-up PR?

jrhea (approved):
Great work!

crates/hyperdrive-math/src/short/max.rs
dpaiton: good call https://github.com/delvtech/hyperdrive-rs/issues/200

crates/hyperdrive-math/src/short/max.rs
dpaiton: Actually I can delete this TODO. The two adjustments we make in the `Err` conditions mean that we are not truly doing Newton's Method, and thus the loss is no longer guaranteed to decrease every step.

PR #202, opened 2024-10-16
calc abs max short guess

dpaiton:
improves the function for calculating the absolute max short.

We use a linear approximation that passes the test 99.67% of the time (7441/7465 passed). There is an additional loop to ensure the remaining tests pass. Future work should be to fix the linear estimate to remove this loop -- it should not be required if the theory & code are correct.

Additional changes:
- removed the solidity comparison test for `calculate_absolute_max_short`. This was failing in cases when solidity was successful and erroneous. The new (rust) starting point obtained by the approximate estimate can succeed when the old (solidity) function fails, and it can also result in a lower final value than the solidity version. Follow-up PRs will add more tests that concretely demonstrate tolerances for final solvency after opening the max short. Once we are happy with the Rust version, we can port the code back over to Solidity & add back the parity tests.
- improved docstrings & comments for format & variable consistency
- renamed `calculate_spot_price` to `calculate_spot_price_down` and added a `*_up` version -- this is helpful for over/under estimating the price in the approximation code.
- made functions for calculating the minimum share reserves and the total fees paid when opening a short to dry up duplicate versions.
- added a new check and cleaned up error messaging/comments in `solvency_after_short`
- modified final liquidity provision in preamble to ensure short is possible given exposure from earlier random trades.
- modified some tests ensure a short is possible before conducting the test

## Derivation of approximate function

image
image
image
image
image

### DIFF
--- a/.github/workflows/rust_test.yml
+++ b/.github/workflows/rust_test.yml
@@ -39,7 +39,7 @@ jobs:
 
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2024-05-02
+          toolchain: nightly
           override: true
           components: rustfmt, clippy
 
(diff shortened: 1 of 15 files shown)


jrhea (approved):
great work!

PR #203, opened 2024-11-26
update to v0.20.0; support hyperdrive v1.0.20

dpaiton:
Updates generated wrappers & version bump

The update has caused `fuzz_sol_calculate_open_short` to fail, possibly due to initialization value changes. I increased the tolerance from `0` to `1e8`.


jalextowle (approved):
LGTM

PR #206, opened 2024-12-04
allow negative solvency after short derivative; add fuzz test

dpaiton:
allow negative gradients & add a test.

### DIFF
--- a/crates/hyperdrive-math/src/lib.rs
+++ b/crates/hyperdrive-math/src/lib.rs
@@ -356,7 +356,7 @@ mod tests {
 
     #[tokio::test]
     async fn fuzz_reserves_given_rate_ignoring_exposure() -> Result<()> {
-        let test_tolerance = fixed!(1e15);
+        let test_tolerance = fixed!(1e16);
         let mut rng = thread_rng();
         let mut counter = 0;
         for _ in 0..*FAST_FUZZ_RUNS {
--- a/crates/hyperdrive-math/src/short/max.rs
+++ b/crates/hyperdrive-math/src/short/max.rs
@@ -546,8 +546,8 @@ impl State {
             // Calculate the next iteration of Newton's method. If the candidate
             // is larger than the absolute max, we've gone too far and something
             // has gone wrong.
-            let derivative = match self.solvency_after_short_derivative_negation(max_bond_guess) {
-                Ok(derivative) => derivative,
+            let derivative = match self.solvency_after_short_derivative(max_bond_guess) {
+                Ok(derivative) => derivative.change_type::<U256>()?,
                 Err(_) => break,
             };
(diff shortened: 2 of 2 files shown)


crates/hyperdrive-math/src/test_utils/preamble.rs
```diff
@@ -72,8 +72,16 @@ pub async fn initialize_pool_with_random_state(
...
     // Add some liquidity again to make sure future bots can make trades.
     let liquidity_amount = rng.gen_range(fixed!(1_000e18)..=fixed!(100_000_000e18));
-    alice.fund(liquidity_amount).await?;
-    alice.add_liquidity(liquidity_amount, None).await?;
+    let exposure = FixedPoint::try_from(
+        alice
+            .get_checkpoint_exposure(alice.latest_checkpoint().await?)
+            .await?
+            .max(I256::zero()),
+    )? + alice.get_state().await?.long_exposure();
```
jalextowle, line 80: This is a bit confusing to me. I feel like we're double counting the chekpoint exposure. Shouldn't this already be in the long exposure?

dpaiton: Honestly my understanding of exposure is shaky. I thought the checkpoint exposure and long exposure were different values.

The reason for this new code is to make it more likely that a short is possible after the preamble. The `solvency_after_short` fn has this check:

```rs
        let exposure_shares = {
            let checkpoint_exposure = FixedPoint::try_from(checkpoint_exposure.max(I256::zero()))?;
            if self.long_exposure() < checkpoint_exposure {
                return Err(eyre!(
                    "expected long_exposure={:#?} >= checkpoint_exposure={:#?}.",
                    self.long_exposure(),
                    checkpoint_exposure
                ));
            } else {
                (self.long_exposure() - checkpoint_exposure).div_up(self.vault_share_price())
            }
        };
        if new_share_reserves >= exposure_shares + self.minimum_share_reserves() {
            Ok(new_share_reserves - exposure_shares - self.minimum_share_reserves())
        } else {
            Err(eyre!("Insiffucient liquidity. Expected share_reserves={:#?} >- exposure_shares={:#?} + min_share_reserves={:#?}",
            new_share_reserves, exposure_shares, self.minimum_share_reserves()))
        }
```

The addition to the preamble is to ensure this is always true by adding it to the final liquidity amount.

jalextowle: Long exposure is the sum of all of the positive checkpoint exposures

dpaiton: it's the sum of all _previous_ positive checkpoint exposures, right? or does it include the current checkpoint?

dpaiton: I removed these lines in #202

PR #207, opened 2024-12-18
hyperdrive 0.20.1; fix checkpoint exposure bug

dpaiton:
A Hyperdrive update from this time last year (https://github.com/delvtech/hyperdrive/pull/692) reworked exposure such that the state `long_exposure` variable included exposure adjustments from any current trade, without representing negative exposure. For example, [here](https://github.com/delvtech/hyperdrive/blob/02d3cd465afbf86cc5f490c618a3a850c6dc6bfe/contracts/src/internal/HyperdriveShort.sol#L403) is the latest code for long exposure in open short. This update did not get fully propagated to the short side of the rust SDK. The long side  can remain unchanged since negative checkpoint exposure can net out future longs, which is not accounted for in the `state.long_exposure` variable.

This PR implements the fix. I will do a round of updating test tolerances once the remaining max short fixes are implemented.

The commits are organized to separate meaningful changes from regex propagation.


crates/hyperdrive-math/src/short/max.rs
```diff
@@ -825,15 +806,16 @@ impl State {
             * (fixed!(1e18) - self.governance_lp_fee())
             / self.vault_share_price())
         .change_type::<I256>()?;
-        Ok(-(lhs - rhs))
+        let exposure_prime = (bond_amount / self.vault_share_price()).change_type::<I256>()?;
```
dpaiton: this is incorrect; fixed in a later commit

crates/hyperdrive-math/src/long/max.rs
```diff
@@ -326,11 +323,8 @@ impl State {
         &self,
         estimate_price: FixedPoint<U256>,
         spot_price: FixedPoint<U256>,
-        checkpoint_exposure: I256,
     ) -> Result<FixedPoint<U256>> {
-        let checkpoint_exposure = FixedPoint::try_from(-checkpoint_exposure.min(int256!(0)))?;
-        let mut estimate =
-            self.calculate_solvency()? + checkpoint_exposure / self.vault_share_price();
+        let mut estimate = self.calculate_solvency()? / self.vault_share_price();
```
dpaiton, line 319: please verify this -- the long exposure should be accounted for elsewhere, so we don't add it here.

dpaiton:
Per conversations with @jalextowle, the checkpoint exposure is ok to leave alone in the Long flow. This PR only updates the short flow to avoid double counting exposure.

sentilesdal (approved):
Looks good to me.

PR #208, opened 2024-12-20
add test to verify inversion of base/bonds in open short

dpaiton:
# Resolved Issues
https://github.com/delvtech/hyperdrive-rs/issues/185

# Description
This PR adds a test to verify the conversion between base and bonds via open short functions is consistent.
The test follows the recommendation in the cited issues.
The test currently FAILS, so it is ignored and a follow-up PR will implement fixes so that it passes.


sentilesdal (approved):
lgtm

PR #209, opened 2025-01-10
fixes for max short test

dpaiton:
# Description
The primary contributions in this PR are:
1. update the minimum share reserves after short from `z_min + max(zeta, 0) + e/c` to `z_min + max(zeta, e/c, 0)`. This update is justified because the checks are performed independently, and thus the reserves do not need to account for share reserves **plus** exposure. Just whichever is larger.
2. update the approximate absolute max short to account for the fact that opening a short reduces exposure. Here is the updated logic:

image
image

## Other changes:
- updated a handful of functions to return `FixedPoint::<I256>` instead of `FixedPoint::<U256>`. This cleaned up the logic in the max short functions.
- `fuzz_sol_calculate_max_long` gets a tolerance -- it was intermittently failing in PRs. Later we should fix this, as requested in https://github.com/delvtech/hyperdrive-rs/issues/136

### DIFF
--- a/bindings/hyperdrivepy/src/hyperdrive_state_methods/short/open.rs
+++ b/bindings/hyperdrivepy/src/hyperdrive_state_methods/short/open.rs
@@ -54,8 +54,11 @@ impl HyperdriveState {
                     err
                 ))
             })?;
-
-        let result = U256::from(result_fp).to_string();
+        let result = U256::try_from(result_fp)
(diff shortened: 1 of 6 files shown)


sentilesdal (approved):
went through this PR in detail IRL with @dpaiton.  Looks good, nothing broken, better than it was before.  Approved.

PR #211, opened 2025-01-10
cleanup tests & test support code

dpaiton:
# Description
- I moved a validity check from `calculate_open_short` to `solvency_after_short`. And we now call `solvency_after_short` in `calculate_open_short`. This diverges a bit from Solidity because we do not split the `apply_open_short` function from the `calculate_open_short` function in rust. This means it was possible to get a valid answer from `calculate_open_short` that would revert in the contract. That should be fixed, now.
- I updated a bunch of tests to check if a short is possible by directly calling `solvency_after_short(min_txn)` instead of the incorrect check it was doing.
- I removed the `get_max_short` function from the preamble. We don't need all of the extra checks.
- I modified the preamble random state generator to be faster by taking fewer steps and using the approximate max short. This was a really slow step in the solidity tests.
- the Agent class now uses the new functions when  you call `agent.calculate_max_short`
- I increased the defaults for `calculate_short_bonds_given_deposit` to `maybe_base_tolerance=1e10` (from `1e9`) and `maybe_max_iterations=1_000` (from `500`).


dpaiton:
I was able to fix the failing test locally by increasing the budget tolerance from `1e9` to `1e10` in `crates/hyperdrive-math/src/short/max.rs:fuzz_calculate_max_short_budget_consumed` (line 1205).

I've lost write privileges to these branches. @sentilesdal would you mind pushing up a commit with that fix to see if CI passes after?

sentilesdal (approved):
Went through with Dylan IRL.  This is basically just test cleanups right before he left.  Approved.

dpaiton:
Lingering issues are summarized here https://github.com/delvtech/hyperdrive-rs/issues/136#issuecomment-2608480255

